PR #6, opened 2021-05-28
feat: Basic auth for projects

HideBa:
## What I've made change
* Added the fields(`isBasicAuthActive`, `basicAuthUsername`, `basicAuthPassword`) on GraphQL schema and domain object
* Implemented the usecase layer
* Save data on the database

## Which part especially I want you to review
* The basic architectural file structure
* Any other files I need to update

## How I tested
* Wrote test code for pkg
* Checked that database saves the data correctly
* Checked the mutation is working by using GQL playground
* Checked the mutation and query is fine by calling API from front-end


schema.graphql
```diff
@@ -179,6 +179,9 @@ enum Role {
 type Project implements Node {
     id: ID!
     isArchived: Boolean!
+    isBasicAuthActive: Boolean!
+    basicAuthUsername: String
+    basicAuthPassword: String
```
rot1024, line 184: Username/Passwordは `String!` でも問題ありません（空文字になる）

rot1024 (commented):
OKです。あとは実際に認証をかける部分ですね。

pkg/project/project_test.go
```diff
@@ -268,3 +268,97 @@ func TestProject_PublicName(t *testing.T) {
...
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			res := tc.p.BasicAuthUsername()
+			assert.Equal(t, tc.expected, res)
+		})
+	}
+}
```
rot1024, line 346: Since there is only one test case, you can write a simpler test instead of a table-driven test.

rot1024 (approved):

KaWaite (approved):

PR #8, opened 2021-06-01
feat: add endpoint to serve basic auth data

HideBa:
## What
* Added the endpoint to serve the basic auth data of the project

## Which part especially should be reviewed
* If there are any problems by adding the new endpoint


internal/app/app.go
```diff
@@ -140,6 +141,25 @@ func apiPublished(cfg *ServerConfig) echo.HandlerFunc {
 	}
 }
 
+func apiBasicAuth(cfg *ServerConfig) echo.HandlerFunc {
```
HideBa, line 144: I created this PR's branch from another branch which approved by @rot1024  but not by @KaWaite . I expect the branch will be merged soon. The difference is only this file.

HideBa:
This change had been merge as a part of #8

PR #11, opened 2021-06-04
feat: published page api

rot1024:
This PR adds new APIs:

- `GET /p/:name/`: return index.html for published page (basic auth supported). HTML file will be fetched from the URL specified by `REEARTH_PUBLISHED_INDEXURL` (defaults to `published.html` in the local `web` directory)
- `GET /p/:name/data.json`: return data.json for published page (basic auth supported).

This change is especially important for the all-in-one Docker image (`reearth/reearth`), as it provides a way to browse published projects without any additional application.

Also I did some refactoring in `internal/app`.


internal/app/public.go
```diff
@@ -1,109 +1,145 @@
...
+	contr := http1.NewPublishedController(i)
+
+	key := struct{}{}
+	auth := middleware.BasicAuthWithConfig(middleware.BasicAuthConfig{
+		Validator: func(user string, password string, c echo.Context) (bool, error) {
+			md, ok := c.Request().Context().Value(key).(interfaces.ProjectPublishedMetadata)
+			if !ok {
+				return true, echo.ErrNotFound
+			}
+			return !md.IsBasicAuthActive || user == md.BasicAuthUsername && password == md.BasicAuthPassword, nil
```
HideBa, line 103: This is not a critical problem but not to be guessed values from the time to compare, it's recommended to use [ConstantTimeCompare](https://echo.labstack.com/middleware/basic-auth/) to compare values.

HideBa (approved):
Just left one commnet. Basically looks good to me. 👍

internal/app/public.go
rot1024: Great. Thank you.

KaWaite (approved):

HideBa (approved):

HideBa:
@rot1024 Do you forget to merge this PR or on purpose? If it's ready, plz merge.

rot1024:
似たようなロジックが分散しちゃうと、一方を変えて他方変え忘れた他とか発生するんですよね。今回のバグはその良い例かな？

PR #17, opened 2021-06-28
fix: ogp image for published page

yk-eukarya:
# Overview

## What I've done
- return project image for OPG if public image does not exists
- fix routing 

## How I tested
Tested by calling the api manually


mimoham24 (approved):
I'm not sure about the logic please ask @KaWaite or @HideBa , but technically looks good

internal/usecase/interfaces/published.go
```diff
@@ -19,10 +19,23 @@ type ProjectPublishedMetadata struct {
 }
 
 func ProjectPublishedMetadataFrom(prj *project.Project) ProjectPublishedMetadata {
```
HideBa, line 21: Is there any reason that this function is declared here since this file usually has only type and interfaces. I assume this function can be moved from this file to `interactor/published.go`. 
@rot1024 How do you think about this?

HideBa (commented):
Looks good except where I left the comment. 👍

internal/usecase/interfaces/published.go
rot1024: No problem.

`ProjectPublishedMetadata` is defined at `interfaces` package. `ProjectPublishedMetadataFrom` is operation using ProjectPublishedMetadata and does not have any dependency on other uppers layers. That's why it can be treated as a `interfaces`.

internal/usecase/interfaces/published.go
```diff
@@ -19,10 +19,23 @@ type ProjectPublishedMetadata struct {
...
+	if title == "" {
+		title = prj.Name()
+	}
+	if description == "" {
+		description = prj.Description()
+	}
+	if image == "" {
+		image = prj.ImageURL().String()
+	}
+
```
rot1024, line 34: Whether this change is appropriate is up for debate; the reason we wanted to have a field named public was to accommodate the need for editors to keep their administrative names and images separate from their public names and images.

Rather than making this change here, it might be better to change the public field to change when the name or image is changed on the frontend side.

@HideBa How about it?

PR #18, opened 2021-07-01
fix: dataset link merge bug #378

yk-eukarya:
# Overview
When a dataset child layer is linked and overrides that link, the value isn't correctly merged.

## What I've done
Returned the overridden value if its overridden

## How I tested
By writing a unit test to test that case

Not sure if the solution may effect the other scenarios

### DIFF
--- a/internal/graphql/resolver_property.go
+++ b/internal/graphql/resolver_property.go
@@ -470,7 +470,7 @@ func actualValue(datasetLoader dataloader.DatasetDataLoader, value interface{}, 
 		return &value, nil
 	}
 	// 先頭のリンクにしかDatasetが割り当てられていない→先頭から順々に辿っていく
-	if len(links) >= 1 && links[0].DatasetID != nil && links[len(links)-1].DatasetID == nil {
+	if len(links) > 1 && links[0].DatasetID != nil && links[len(links)-1].DatasetID == nil {
 		dsid := *links[0].DatasetID
 		for i, link := range links {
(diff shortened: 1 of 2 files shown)


internal/graphql/resolver_property.go
```diff
@@ -465,12 +465,12 @@ func (*propertyGroupResolver) SchemaGroup(ctx context.Context, obj *graphql1.Pro
...
 
-func actualValue(datasetLoader dataloader.DatasetDataLoader, value interface{}, links []*graphql1.PropertyFieldLink) (interface{}, error) {
-	if len(links) == 0 {
+func actualValue(datasetLoader dataloader.DatasetDataLoader, value interface{}, links []*graphql1.PropertyFieldLink, overridden bool) (interface{}, error) {
+	if len(links) == 0 || overridden {
 		return &value, nil
 	}
 	// 先頭のリンクにしかDatasetが割り当てられていない→先頭から順々に辿っていく
-	if len(links) > 1 && links[0].DatasetID != nil && links[len(links)-1].DatasetID == nil {
+	if len(links) >= 1 && links[0].DatasetID != nil && links[len(links)-1].DatasetID == nil {
```
rot1024, line 473: If there was only one link, `else` block would be executed and it should work. Did you change this because it didn't work?

rot1024 (changes requested):

internal/graphql/resolver_property.go
yk-eukarya: Sorry about this, It seems I forgot it while debugging

PR #19, opened 2021-07-05
feat: widget align system

KaWaite:
# Overview
- Add WidgetAlignSystem to backend to handle how the widgets will be placed in the Editor on the front-end.
- Add WidgetLayout to Plugin's extensions to handle how the widget will be handled with the align system.
- Includes migrations for both plugins(adding the WidgetLayout to all plugin's extensions) and scene (adding a widget align system to all already made scenes)

## What I've done
- Develop a widget align system to handle alignment of enabled widgets
- Added WidgetLayout to Plugin/sceneWidget
- Added WidgetAlignSystem to scene
- DB migration for adding alignSystem to scenes

## How I tested
- with graphql queries and mutations, in graphql playground, added, removed, updated widget align system
- tested both migrations on local machine

- Everything


schema.graphql
```diff
@@ -826,12 +899,14 @@ input UpdateWidgetInput {
...
   extensionId: PluginExtensionID!
   enabled: Boolean
+  layout: WidgetLayoutInput
 }
 
 input RemoveWidgetInput {
   sceneId: ID!
   pluginId: PluginID!
   extensionId: PluginExtensionID!
+  location: LocationInput!
```
rot1024, line 909: This is not needed because we can identify specific widget with only one widget ID

schema.graphql
```diff
@@ -815,6 +873,21 @@ input DeleteProjectInput {
...
 
+input LocationInput {
+  zone: String!
+  section: String!
+  area: String!
+}
+
+input WidgetLayoutInput {
+  extended: Boolean
+	Location: LocationInput!
```
rot1024, line 884: Not needed because there are no duplicated widget id so we can identify specific widget by one id

schema.graphql
```diff
@@ -276,6 +308,30 @@ type PluginExtension {
...
+}
+
+type WidgetSection {
+  top: WidgetArea
+  middle: WidgetArea
+  bottom: WidgetArea
+}
+
+type WidgetArea {
+  widgetIds: [ID]
```
rot1024, line 331: Should be `[ID!]!`

schema.graphql
```diff
@@ -250,6 +250,36 @@ type PluginMetadata {
...
+  RIGHT
+}
+
+enum Area {
+  TOP
+  MIDDLE
+  BOTTOM
+}
+
+type Location {
```
rot1024, line 270: Rename it to WidgetLocation, otherwise Location is too abstract and confusing because it sounds like LatLng

internal/adapter/graphql/controller_scene.go
```diff
@@ -70,10 +94,18 @@ func (c *SceneController) UpdateWidget(ctx context.Context, ginput *UpdateWidget
 }
 
 func (c *SceneController) RemoveWidget(ctx context.Context, ginput *RemoveWidgetInput, operator *usecase.Operator) (*RemoveWidgetPayload, error) {
+	var loc scene.Location
+	if ginput.Location != nil {
+		l := ginput.Location
```
rot1024, line 99: Remember this to shorten codes:

```
if a := b; a != nil {

}
```

internal/adapter/graphql/controller_scene.go
```diff
@@ -70,10 +94,18 @@ func (c *SceneController) UpdateWidget(ctx context.Context, ginput *UpdateWidget
 }
 
 func (c *SceneController) RemoveWidget(ctx context.Context, ginput *RemoveWidgetInput, operator *usecase.Operator) (*RemoveWidgetPayload, error) {
+	var loc scene.Location
```
rot1024, line 97: If ginput.Layout is nil, this should be also nil

```
var loc *scene.Location
```

internal/infrastructure/mongo/mongodoc/plugin.go
```diff
@@ -3,17 +3,19 @@ package mongodoc
...
-	Schema      string
-	Visualizer  string
+	ID           string
+	Type         string
+	Name         map[string]string
+	Description  map[string]string
+	Icon         string
+	Schema       string
+	Visualizer   string
+	WidgetLayout *scene.WidgetLayout
```
rot1024, line 18: Don't include domain model in  mongodoc fields

internal/infrastructure/mongo/mongodoc/scene_alignsystem.go
```diff
@@ -0,0 +1,190 @@
...
+				Align:     *was.Alignment("inner", "left", "top"),
+			},
+			Middle: WidgetArea{
+				WidgetIds: toString(was.WidgetIds("inner", "left", "middle")),
+				Align:     *was.Alignment("inner", "left", "middle"),
+			},
+			Bottom: WidgetArea{
+				WidgetIds: toString(was.WidgetIds("inner", "left", "bottom")),
+				Align:     *was.Alignment("inner", "left", "bottom"),
+			},
```
rot1024, line 57: Don't repeat yourself (Section is repeated)

rot1024 (changes requested):
Reviewed partially

pkg/plugin/extension_builder_test.go
```diff
@@ -75,6 +84,7 @@ func TestExtensionBuilder_Build(t *testing.T) {
 			description:   i18n.StringFrom("ddd"),
 			schema:        id.MustPropertySchemaID("foo#1.1.1/hhh"),
 			visualizer:    "vvv",
+			widgetLayout:  nil,
```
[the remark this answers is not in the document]
KaWaite: This is for the widget's layout, not the alignment system. But I will change it so it isn't nil!

pkg/plugin/extension_test.go
```diff
@@ -34,6 +37,7 @@ func TestExtension(t *testing.T) {
 		Description(i18n.StringFrom("ddd")).
 		Schema(id.MustPropertySchemaID("hoge#0.1.0/fff")).
 		Icon("test").
+		WidgetLayout(nil).
```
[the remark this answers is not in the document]
KaWaite: This is also to do with layout, not the alignment system. But again, I will make it not nil to check that the method is returning a widget layout.

pkg/scene/builder.go
```diff
@@ -85,6 +88,13 @@ func (b *Builder) WidgetSystem(widgetSystem *WidgetSystem) *Builder {
 	return b
 }
 
+// WidgetAlignSystem _
```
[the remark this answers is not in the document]
KaWaite: Sorry just to be clear, you want ALL the `// xxx _` comments to be removed from this file?

schema.graphql
KaWaite: The location is used to remove the widgetID from the align system. Without it, I need to create a function that will retrieve each WidgetArea and search its widgetIds until it finds the correct widgetId.
I thought that would be heavy so decided knowing the location was better.
How would you prefer this to be handled?

pkg/scene/builder.go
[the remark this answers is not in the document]
rot1024: Yes, but it is hard to edit many files at once because many of them have such comments. Therefore, it would be nice if you could delete the comments left in the file when you edit it.

Incidentally, the reason why such comments are still there is that before the introduction of golangci-lint, all types and functions had to be commented or else a warning would appear.

schema.graphql
rot1024: That's not a problem because it's not a very heavy process. The information in the widget alignment system is always read and written together with the Scene, and all the information is put into memory at once. Their retrieval speed is usually not a concern at all (unless, for example, the process of scanning all the millions of records is repeated several times a second).

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	a := was.WidgetArea(l.Zone, l.Section, l.Area)
+
+	var nwid []*id.WidgetID
+	for _, w := range a.widgetIds {
+		if w.ID() != wid.ID() {
+			nwid = append(nwid, w)
+		}
+	}
```
rot1024, line 137: Instead of searching for widgets by location, search for widgets for all locations from the beginning. Otherwise, the location will behave as an ID.

In general, the information that behaves as an ID should be kept to a minimum, otherwise it will cause human errors and bugs to be introduced.

By the way, with this amount of data, you don't need to worry about the search speed at all.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+	case "inner":
+		return &was.inner
+	case "outer":
+		return &was.outer
+	}
+	return nil
+}
+
+// WidgetSection will return a specific section in the align system.
+func (was *WidgetAlignSystem) WidgetSection(zone, section string) *WidgetSection {
```
rot1024, line 59:
```suggestion
func (was *WidgetAlignSystem) Section(zone, section string) *WidgetSection {
```

You can rather shorten the name here, since the only possible Section of WidgetAlignSystem is widget's one, so let's keep the method name clear.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+	Area    string
+}
+
+// NewWidgetAlignSystem returns a new widget align system.
+func NewWidgetAlignSystem() *WidgetAlignSystem {
+	return &WidgetAlignSystem{}
+}
+
+// WidgetZone will return a specific zone in the align system.
+func (was *WidgetAlignSystem) WidgetZone(zone string) *WidgetZone {
```
rot1024, line 45:
```suggestion
func (was *WidgetAlignSystem) Zone(zone string) *WidgetZone {
```

You can rather shorten the name here, since the only possible Zone of WidgetAlignSystem is widget's one, so let's keep the method name clear.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+	case "center":
+		return &z.center
+	case "right":
+		return &z.right
+	}
+	return nil
+}
+
+// WidgetArea will return a specific area in the align system.
+func (was *WidgetAlignSystem) WidgetArea(zone, section, area string) *WidgetArea {
```
rot1024, line 78:
```suggestion
func (was *WidgetAlignSystem) Area(zone, section, area string) *WidgetArea {
```

You can rather shorten the name here, since the only possible Area of WidgetAlignSystem is widget's one, so let's keep the method name clear.

pkg/scene/builder/scene.go
```diff
@@ -62,11 +65,19 @@ func (b *Builder) widgets(ctx context.Context, s *scene.Scene, p []*property.Pro
 			PluginID:    w.Plugin().String(),
 			ExtensionID: string(w.Extension()),
 			Property:    b.property(ctx, findProperty(p, w.Property())),
+			Extended:    w.WidgetLayout().Extended,
 		})
 	}
 	return res
 }
 
+func (b *Builder) widgetAlignment(ctx context.Context, s *scene.Scene) *widgetAlignSystemJSON {
```
rot1024, line 74:
```suggestion
func (b *Builder) widgetAlignSystem(ctx context.Context, s *scene.Scene) *widgetAlignSystemJSON {
```

The names should be consistent.

pkg/scene/builder/builder_test.go
```diff
@@ -380,8 +380,8 @@ func TestSceneBuilder(t *testing.T) {
 
 	sceneWidgetID1 := id.NewWidgetID()
 	sceneWidgetID2 := id.NewWidgetID()
-	sceneWidget1 := scene.MustNewWidget(&sceneWidgetID1, pluginID, pluginExtension1ID, scenePropertyID, false)
-	sceneWidget2 := scene.MustNewWidget(&sceneWidgetID2, pluginID, pluginExtension2ID, scenePropertyID, true)
+	sceneWidget1 := scene.MustNewWidget(&sceneWidgetID1, pluginID, pluginExtension1ID, scenePropertyID, false, nil)
+	sceneWidget2 := scene.MustNewWidget(&sceneWidgetID2, pluginID, pluginExtension2ID, scenePropertyID, true, nil)
```
rot1024, line 384: You should test the build of WidgetAlignSystem. In particular, testing here is an essential part of maintaining the quality of the scene's build functionality so that it does not rely on manual test.

pkg/plugin/manifest/parser_test.go
```diff
@@ -28,6 +29,7 @@ var normalExpected = &Manifest{
 		plugin.NewExtension().ID(id.PluginExtensionID("hoge")).
 			Visualizer(visualizer.VisualizerCesium).
 			Type(plugin.ExtensionTypePrimitive).
+			WidgetLayout(&scene.WidgetLayout{Extendable: true, Extended: false, Floating: false, DefaultLocation: &scene.WidgetLocation{Zone: "outer", Section: "left", Area: "top"}}).
```
rot1024, line 32: One line is long, so please break the line.

pkg/plugin/manifest/convert.go
```diff
@@ -129,6 +133,35 @@ func (i Extension) extension(pluginID id.PluginID, sys bool) (*plugin.Extension,
 	return ext, schema, nil
 }
 
+func (l *WidgetLayout) layout() *scene.WidgetLayout {
+	if l == nil {
+		return &scene.WidgetLayout{Extended: false, Extendable: true, Floating: false, DefaultLocation: &scene.WidgetLocation{Zone: "outer", Section: "left", Area: "top"}}
```
rot1024, line 138: One line is long, so please break the line.

pkg/plugin/extension_test.go
```diff
@@ -5,27 +5,30 @@ import (
...
-		Schema:      id.MustPropertySchemaID("hoge#0.1.0/fff"),
-		Visualizer:  "vvv",
+		Id:           "xxx",
+		Type:         ExtensionTypePrimitive,
+		Name:         i18n.StringFrom("aaa"),
+		Description:  i18n.StringFrom("ddd"),
+		Icon:         "test",
+		Schema:       id.MustPropertySchemaID("hoge#0.1.0/fff"),
+		Visualizer:   "vvv",
+		WidgetLayout: &scene.WidgetLayout{},
```
rot1024, line 31: As I mentioned in another comment, don't empty it, fill it with something inside.

pkg/plugin/extension_builder_test.go
```diff
@@ -75,6 +84,7 @@ func TestExtensionBuilder_Build(t *testing.T) {
 			description:   i18n.StringFrom("ddd"),
 			schema:        id.MustPropertySchemaID("foo#1.1.1/hhh"),
 			visualizer:    "vvv",
+			widgetLayout:  &scene.WidgetLayout{},
```
rot1024, line 87: As I mentioned in another comment, don't empty it, fill it with something inside.

pkg/plugin/extension_builder_test.go
```diff
@@ -51,6 +52,13 @@ func TestExtensionBuilder_Visualizer(t *testing.T) {
 	res := b.ID("xxx").Visualizer("ccc").MustBuild()
 	assert.Equal(t, visualizer.Visualizer("ccc"), res.Visualizer())
 }
+func TestExtensionBuilder_WidgetLayout(t *testing.T) {
+	var b = NewExtension()
+	wl := &scene.WidgetLayout{Extendable: true}
+	wl2 := &scene.WidgetLayout{Extendable: true}
+	res := b.ID("xxx").WidgetLayout(wl).MustBuild()
+	assert.Equal(t, wl2, res.WidgetLayout())
```
rot1024, line 60:
```suggestion
	assert.Same(t, wl2, res.WidgetLayout())
```

Use Equal and Same as needed; Same is needed for more stringent testing of identities; Equal will not detect them.

pkg/plugin/extension.go
```diff
@@ -72,6 +74,11 @@ func (w *Extension) Visualizer() visualizer.Visualizer {
 	return w.visualizer
 }
 
+// WidgetMetaData _
+func (w *Extension) WidgetLayout() *scene.WidgetLayout {
```
rot1024, line 78:
```suggestion
func (w *Extension) Layout() *scene.WidgetLayout {
```

`widget.WidgetLayout()`, the widget is repeated twice. `widget.Layout()` conveys the meaning well enough.

internal/usecase/interfaces/scene.go
```diff
@@ -29,7 +39,7 @@ type Scene interface {
 	Create(context.Context, id.ProjectID, *usecase.Operator) (*scene.Scene, error)
 	AddWidget(context.Context, id.SceneID, id.PluginID, id.PluginExtensionID, *usecase.Operator) (*scene.Scene, *scene.Widget, error)
 	UpdateWidget(context.Context, UpdateWidgetParam, *usecase.Operator) (*scene.Scene, *scene.Widget, error)
-	RemoveWidget(context.Context, id.SceneID, id.PluginID, id.PluginExtensionID, *usecase.Operator) (*scene.Scene, error)
+	RemoveWidget(context.Context, id.SceneID, id.PluginID, id.PluginExtensionID, *scene.WidgetLocation, *usecase.Operator) (*scene.Scene, error)
```
rot1024, line 42: If you follow the review comments I wrote in the GraphQL schema, you will need to fix them as well.

internal/usecase/interfaces/scene.go
```diff
@@ -9,11 +9,21 @@ import (
...
 )
 
+type LayoutParams struct {
+	Extended *bool
+	Location,
+	NewLocation *scene.WidgetLocation
+	OldIndex,
+	NewIndex *int
+	Align *string
+}
```
rot1024, line 19: If you follow the review comments I wrote in the GraphQL schema, you will need to fix this as well.

internal/usecase/interactor/scene.go
```diff
@@ -193,12 +193,17 @@ func (i *Scene) AddWidget(ctx context.Context, id id.SceneID, pid id.PluginID, e
...
+
+	widget, err = scene.NewWidget(nil, pid, eid, property.ID(), true, &scene.WidgetLayout{Extendable: widgetLayout.Extendable, Extended: widgetLayout.Extended, Floating: widgetLayout.Floating, DefaultLocation: widgetLayout.DefaultLocation})
 	if err != nil {
 		return nil, nil, err
 	}
 
 	s.WidgetSystem().Add(widget)
+	if !widget.WidgetLayout().Floating {
+		s.WidgetAlignSystem().Add(widget.ID().Ref(), widget.WidgetLayout().DefaultLocation)
+	}
```
rot1024, line 206: This comment is more developmental advice. If you have time, try it.

Currently, the scene has a WidgetSystem, which is a map of widget IDs and properties, and a WidgetAlignSystem. This means that we now have to describe in the use case layer the changes we want to happen in the WidgetAlignSystem when we add a widget to the WidgetSystem. This requires such "special" knowledge from the implementor, that when you forget to implement it, you get a bug.

Therefore, incorporating the WidgetAlignSystem in the WidgetSystem makes the code safer and less bug-prone, since this process can be hidden in the domain layer. The use case layer just needs to call widgetSystem.Add and everything is done.

internal/infrastructure/mongo/migration/210730152508_add_widget_layout.go
```diff
@@ -0,0 +1,55 @@
...
+					if w.WidgetLayout == nil {
+						pid, _ := id.PluginIDFrom(w.Plugin)
+						wl := builtin.GetPlugin(pid).Extension(id.PluginExtensionID(w.Extension)).WidgetLayout()
+						float := wl.Floating
+						extendable := wl.Extendable
+						extended := false
+						dl := wl.DefaultLocation
+						mdl := mongodoc.WidgetLocation{Zone: dl.Zone, Section: dl.Section, Area: dl.Area}
+						wldoc := mongodoc.WidgetLayout{Extendable: extendable, Extended: extended, Floating: float, DefaultLocation: &mdl}
+						w.WidgetLayout = &wldoc
```
rot1024, line 42:
```suggestion
						w.WidgetLayout = &mongodoc.WidgetLayout{
							Extendable: wl.Extendable,
							Extended: false,
							Floating: wl.Floating,
							DefaultLocation: &mongodoc.WidgetLocation{
								Zone: dl.Zone,
								Section: wl.DefaultLocation.Section,
								Area: wl.DefaultLocation.Area,
							},
						}
```

Please be proactive in creating new lines. Difficult-to-read code can cause bugs.

Also, placing local variables can sometimes make the code even more difficult to understand.

internal/infrastructure/mongo/mongodoc/scene_alignsystem.go
```diff
@@ -0,0 +1,129 @@
...
+	Center WidgetSection
+	Right  WidgetSection
+}
+
+type SceneAlignSystemDocument struct {
+	Inner WidgetZone
+	Outer WidgetZone
+}
+
+func NewWidgetAlignSystem(was scene.WidgetAlignSystem) *SceneAlignSystemDocument {
```
rot1024, line 43:
```suggestion
func NewWidgetAlignSystem(was *scene.WidgetAlignSystem) *SceneAlignSystemDocument {
```

Since the domain model assumes that the WidgetAlignSystem is handled by reference, this New method should also accept the WidgetAlignSystem by reference. The same goes for the others.

internal/infrastructure/mongo/mongodoc/scene_alignsystem.go
```diff
@@ -0,0 +1,129 @@
...
+	was.WidgetAreaFrom(toStruct(d.Outer.Center.Middle.WidgetIds), d.Outer.Center.Middle.Align, "outer", "center", "middle")
+	was.WidgetAreaFrom(toStruct(d.Outer.Center.Bottom.WidgetIds), d.Outer.Center.Bottom.Align, "outer", "center", "bottom")
+	// Outer Right
+	was.WidgetAreaFrom(toStruct(d.Outer.Right.Top.WidgetIds), d.Outer.Right.Top.Align, "outer", "right", "top")
+	was.WidgetAreaFrom(toStruct(d.Outer.Right.Middle.WidgetIds), d.Outer.Right.Middle.Align, "outer", "right", "middle")
+	was.WidgetAreaFrom(toStruct(d.Outer.Right.Bottom.WidgetIds), d.Outer.Right.Bottom.Align, "outer", "right", "bottom")
+	return was, nil
+}
+
+func toString(wids []*id.WidgetID) []string {
```
rot1024, line 77:
```suggestion
func widgetIDsToStrings(wids []*id.WidgetID) []string {
```

The name is too abstract.

internal/infrastructure/mongo/mongodoc/scene_alignsystem.go
```diff
@@ -0,0 +1,129 @@
...
+	for _, wid := range wids {
+		if wid == nil {
+			continue
+		}
+		docids = append(docids, wid.String())
+	}
+	return docids
+}
+
+func toStruct(wids []string) []*id.WidgetID {
```
rot1024, line 91:
```suggestion
func stringsToWidgetIDs(wids []string) []*id.WidgetID {
```
The name is too abstract.

internal/infrastructure/mongo/mongodoc/scene_alignsystem.go
```diff
@@ -0,0 +1,129 @@
...
+	Inner WidgetZone
+	Outer WidgetZone
+}
+
+func NewWidgetAlignSystem(was scene.WidgetAlignSystem) *SceneAlignSystemDocument {
+	widgetAlignDoc := SceneAlignSystemDocument{Inner: buildWidgetZone(&was), Outer: buildWidgetZone(&was)}
+	return &widgetAlignDoc
+}
+
+func (*SceneAlignSystemDocument) ToModelAlignSystem(d SceneAlignSystemDocument) (*scene.WidgetAlignSystem, error) {
```
rot1024, line 48:
```suggestion
func (*SceneAlignSystemDocument) ToModelAlignSystem(d SceneAlignSystemDocument) *scene.WidgetAlignSystem {
```

The code in this method does not return an error

internal/infrastructure/mongo/mongodoc/plugin.go
```diff
@@ -94,6 +97,7 @@ func (d *PluginDocument) Model() (*plugin.Plugin, error) {
 			Name(d.Name).
 			Description(d.Description).
 			Icon(e.Icon).
+			WidgetLayout(&scene.WidgetLayout{Extendable: e.WidgetLayout.Extendable, Extended: e.WidgetLayout.Extended, Floating: e.WidgetLayout.Floating, DefaultLocation: (*scene.WidgetLocation)(e.WidgetLayout.DefaultLocation)}).
```
rot1024, line 100: New line, please.

internal/infrastructure/mongo/mongodoc/plugin.go
```diff
@@ -54,13 +56,14 @@ func NewPlugin(plugin *plugin.Plugin) (*PluginDocument, string) {
...
-			Schema:      e.Schema().String(),
-			Visualizer:  string(e.Visualizer()),
+			ID:           string(e.ID()),
+			Type:         string(e.Type()),
+			Name:         e.Name(),
+			Description:  e.Description(),
+			Icon:         e.Icon(),
+			Schema:       e.Schema().String(),
+			Visualizer:   string(e.Visualizer()),
+			WidgetLayout: &WidgetLayout{Extendable: e.WidgetLayout().Extendable, Extended: e.WidgetLayout().Extended, Floating: e.WidgetLayout().Floating, DefaultLocation: (*WidgetLocation)(e.WidgetLayout().DefaultLocation)},
```
rot1024, line 66: New line, please.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+	return append(array[:index], append([]*id.WidgetID{value}, array[index:]...)...)
+}
+
+// removeInt is used in moveInt to remove the widgetID from original position(index).
+func removeInt(array []*id.WidgetID, index int) []*id.WidgetID {
+	return append(array[:index], array[index+1:]...)
+}
+
+// WidgetAreaFrom will add a slice of widget ids to a specific area of an align system.
+func (was *WidgetAlignSystem) WidgetAreaFrom(wids []*id.WidgetID, align, z, s, a string) {
```
rot1024, line 181:
```suggestion
func (was *WidgetAlignSystem) AddArea(wids []*id.WidgetID, align, z, s, a string) {
```

A function named From is expected to initialize and return some value, just like the New function.

rot1024 (changes requested):

pkg/scene/builder.go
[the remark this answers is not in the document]
KaWaite: Noted! From now on when I see comments like that I will remove them.

schema.graphql
KaWaite: Ok, I'll create such a function, remove this locationInput and update the RemoveWidget method.

schema.graphql
```diff
@@ -815,6 +873,21 @@ input DeleteProjectInput {
...
+  extended: Boolean
+	Location: WidgetLocationInput!
+	newLocation: WidgetLocationInput
+	oldIndex: Int
+	newIndex: Int
+  align: String
+}
+
 input AddWidgetInput {
   sceneId: ID!
```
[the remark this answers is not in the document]
KaWaite: Since widgets have a default location I thought that wouldn't be needed as a widget would be always added to the default location first. 
Do you think we will have UI in the future where the user somehow specifies where the widget will go when they enable the widget?

schema.graphql
```diff
@@ -826,12 +899,14 @@ input UpdateWidgetInput {
   pluginId: PluginID!
   extensionId: PluginExtensionID!
   enabled: Boolean
+  layout: WidgetLayoutInput
```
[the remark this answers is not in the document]
KaWaite: Sorry, in my mind Widget Location is just the location of the widget in the align system (Zone, Section and Area).
In this case for the WidgetLayoutInput, it contains whether the widget is extended, location data, index data and alignment. So I feel like Location is too narrow for the input fields.
Do you feel WidgetLocationInput is still more appropriate considering the above?

pkg/scene/widget_align_system.go
KaWaite: Ok, I misunderstood the naming system! But this function actually adds multiple widgets to an Area, not adding an Area, so how about renaming it AddMultiple?

KaWaite: But I am using it in mongodoc to take mongo's widget align system's widgetIds and add them to the struct widget align system so I am not sure that AddMultiple is appropriate either.

pkg/scene/builder/builder_test.go
KaWaite: This part is to do with a new Widget, so the nil is to do with the WidgetLayout. I will add that though!

schema.graphql
```diff
@@ -250,6 +250,36 @@ type PluginMetadata {
   createdAt: DateTime!
 }
 
+enum Zone {
```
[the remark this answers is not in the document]
KaWaite: I already have WidgetZone, WidgetSection and WidgetArea types and apparently I can't name enums the same. So I updated them as WZone, WSection and WArea. I know that isn't much clearer so please let me know if you have any other ideas.

schema.graphql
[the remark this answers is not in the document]
rot1024: If I were you, I would add a field that allows you to specify the location from the beginning for that possibility. But it is not mandatory.

schema.graphql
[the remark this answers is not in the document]
rot1024: How about `WidgetZoneType` `WidgetAreaType` `WidgetSectionType` ?

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+	case "middle":
+		return &s.middle
+	case "bottom":
+		return &s.bottom
+	}
+	return nil
+}
+
+// WidgetIds will return a slice of widget ids from a specific area.
+func (was *WidgetAlignSystem) WidgetIds(z, s, a string) []*id.WidgetID {
```
rot1024, line 128:
```suggestion
func (was *WidgetAlignSystem) WidgetIDs(z, s, a string) []*id.WidgetID {
```

In Golang, `Id` in type names should be `ID`

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,200 @@
...
+type WidgetSection struct {
+	top    WidgetArea
+	middle WidgetArea
+	bottom WidgetArea
+}
+
+// WidgetArea has the widgets and alignment information found in each part area of a section.
+type WidgetArea struct {
+	widgetIds []*id.WidgetID
+	align     string
```
rot1024, line 30: It is recommended to define consts

```go
const (
	AlignLeft   = "left"
	AlignCenter = "center"
	AlignRight  = "right"
	ZoneInner   = "inner"
	ZoneOuter   = "outer"
)
```

pkg/scene/widget_align_system.go
rot1024: ok just `AddAll` is good

```suggestion
func (was *WidgetAlignSystem) AddAll(wids []*id.WidgetID, z, s, a string) {
```

Note: align is not used

rot1024 (changes requested):
Additional comments (Haven't checked for new changes yet)

schema.graphql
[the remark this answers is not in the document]
rot1024: Ok I see, but I was confused  because also plugin manifest has `widgetLayout` field for each widgets.

pkg/scene/widget_align_system.go
KaWaite: Sorry, I get the general idea. I am guessing I can use these to somehow make sure align is only one of those, but I am not sure how to do that correctly in Golang (in this particular situation). When I googled around a bit I couldn't find anything helpful.
If you could explain a little how you are suggesting to use these consts that would help alot!

pkg/scene/widget_align_system.go
KaWaite: Ah, missed align. This method is currently used in the conversion from mongodoc types to domain models so it is good to convert align as well :P 
Thank you.

pkg/scene/widget_align_system.go
rot1024: Go's standard library is very useful as an example of good code in Go. For example, net/http package provides each HTTP method name as a const and uses it in various places, making it clear that it is a special constant compared to writing it directly as a string.

https://pkg.go.dev/net/http#pkg-constants

https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/net/http/method.go;l=10

KaWaite: Will start using the standard library as a reference!

internal/usecase/interactor/scene.go
KaWaite: Since we are a little hurting for time right now, I am thinking this could be a good refactor task for me later in the month or next month.

internal/usecase/interfaces/scene.go
KaWaite: Yep, updated all the places affected by changes you have suggested!

pkg/plugin/manifest/convert.go
```diff
@@ -129,6 +133,52 @@ func (i Extension) extension(pluginID id.PluginID, sys bool) (*plugin.Extension,
...
+	} else {
+		f := l.Floating
+		swl.Floating = *f
+	}
+	var dl scene.WidgetLocation
+	if l.DefaultLocation == nil {
+		dl = scene.WidgetLocation{
+			Zone:    "outer",
+			Section: "left",
+			Area:    "top",
```
rot1024, line 169:
```suggestion
			Zone:    scene.ZoneOuter,
			Section: scene.SectionLeft,
			Area:    scene.AreaTop,
```

pkg/plugin/manifest/convert.go
```diff
@@ -129,6 +133,52 @@ func (i Extension) extension(pluginID id.PluginID, sys bool) (*plugin.Extension,
...
+		e := l.Extendable
+		swl.Extendable = *e
+	}
+	if l.Floating == nil {
+		swl.Floating = false
+	} else {
+		f := l.Floating
+		swl.Floating = *f
+	}
+	var dl scene.WidgetLocation
```
rot1024, line 164:
```suggestion
	var dl *scene.WidgetLocation
```

Be careful of diff between reference and vale. This code always generates empty WidgetLocation struct (not nil) and it will be set to scene.

pkg/plugin/manifest/convert.go
```diff
@@ -129,6 +133,52 @@ func (i Extension) extension(pluginID id.PluginID, sys bool) (*plugin.Extension,
...
+			Area:    "top",
+		}
+	} else {
+		dl = scene.WidgetLocation{
+			Zone:    l.DefaultLocation.Zone,
+			Section: l.DefaultLocation.Section,
+			Area:    l.DefaultLocation.Area,
+		}
+	}
+	swl.DefaultLocation = &dl
```
rot1024, line 178:
```suggestion
	if dl == nil {
		swl.DefaultLocation = dl
	}
```

Empty struct always will be set

pkg/plugin/manifest/parser_test.go
```diff
@@ -28,6 +29,16 @@ var normalExpected = &Manifest{
...
 			Visualizer(visualizer.VisualizerCesium).
 			Type(plugin.ExtensionTypePrimitive).
+			WidgetLayout(&scene.WidgetLayout{
+				Extendable: true,
+				Extended:   false,
+				Floating:   false,
+				DefaultLocation: &scene.WidgetLocation{
+					Zone:    "outer",
+					Section: "left",
+					Area:    "top",
```
rot1024, line 39:
```suggestion
					Zone:    scene.ZoneOuter,
					Section: scene.SectionLeft,
					Area:    scene.AreaTop,
```

pkg/scene/builder.go
```diff
@@ -48,57 +48,54 @@ func (b *Builder) MustBuild() *Scene {
...
 func (b *Builder) WidgetSystem(widgetSystem *WidgetSystem) *Builder {
 	widgetSystem2 := *widgetSystem
 	b.scene.widgetSystem = &widgetSystem2
 	return b
 }
 
-// RootLayer _
+func (b *Builder) WidgetAlignSystem(widgetAlignSystem *WidgetAlignSystem) *Builder {
+	widgetAlignSystem2 := *widgetAlignSystem
+	b.scene.widgetAlignSystem = &widgetAlignSystem2
```
rot1024, line 84: You don't need to clone it because the contents of WidgetAlignSystem are hidden by private fields and methods. That's what "encapsulation" (OOP term) is for.

```suggestion
	b.scene.widgetAlignSystem = widgetAlignSystem
```

On the other hand, value objects whose fields are all public should be cloned. Otherwise, unintended scene changes will occur. But even in that case, this code is not safe to clone　when widgetAlignSystem has any references as children.

pkg/scene/widget.go
```diff
@@ -21,17 +29,22 @@ func NewWidget(wid *id.WidgetID, plugin id.PluginID, extension id.PluginExtensio
...
+		property:     property,
+		enabled:      enabled,
+		widgetLayout: widgetLayout,
 	}, nil
 }
 
-func MustNewWidget(wid *id.WidgetID, plugin id.PluginID, extension id.PluginExtensionID, property id.PropertyID, enabled bool) *Widget {
-	w, err := NewWidget(wid, plugin, extension, property, enabled)
+func MustNewWidget(wid *id.WidgetID, plugin id.PluginID, extension id.PluginExtensionID, property id.PropertyID, enabled bool, widgetLayout *WidgetLayout) *Widget {
+	w, err := NewWidget(wid, plugin, extension, property, enabled, widgetLayout)
```
rot1024, line 47: This function has too many parameters, so consider turning it into a builder pattern.

pkg/scene/widget.go
```diff
@@ -58,6 +71,14 @@ func (w *Widget) Enabled() bool {
...
+	return w.widgetLayout
+}
+
 func (w *Widget) SetEnabled(enabled bool) {
 	w.enabled = enabled
 }
+
+func (w *Widget) SetExtended(extended bool) {
+	w.widgetLayout.Extended = extended
+}
```
rot1024, line 84: Not tested

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+type WidgetLocation struct {
+	Zone    string
+	Section string
+	Area    string
+}
+
+const (
+	AlignStart  = "start"
```
rot1024, line 40:
```suggestion
	WidgetAlignStart  = "start"
```

Sorry, but I found that these constants should have `Widget` prefix because when we want to use it in other packages, we write `scene.AlignStart` but we may confuse "what is align for???" If they have the prefix, `scene.WidgetAlignStart` is very clear that the "align" is for widgets.

Other constants should be renamed also.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+	AlignCenter = "center"
+	AlignEnd    = "end"
+)
+
+var Areas = []string{
+	"top",
+	"middle",
+	"bottom",
```
rot1024, line 48:
```suggestion
	AlignTop,
	AlignMiddle,
	AlignBottom,
```

We have all these constants, but they're not being used!

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+	"center": Areas,
+	"right":  Areas,
+}
+
+var Zones = map[string]map[string][]string{
+	"inner": Sections,
+	"outer": Sections,
+}
+
+func (was *WidgetAlignSystem) FindWidgetIDLocation(wid id.WidgetID) (*int, *WidgetLocation) {
```
rot1024, line 62: There is room for reconsideration here.

First, the name could be a bit smarter (FindWidgetIDLocation -> FindWidgetLocation). Second, int doesn't have to be a reference. A zero value will suffice.

And the method itself is not very good. Do we really need to return an index? Why do we need this method?

It is because the logic has been concentrated in the WidgetAlignSystem. If each Area's method is called internally by each Section, then each Zone calls that method internally, and finally the WidgetAlignSystem calls the Zone's method, each would only have to worry about the one below it. However, currently, the WidgetAlignSystem takes care of all Areas, and the contents of the Areas are leaking out.

For example, if we had such a method, each would not have to go to the private fields of other structs.

Message passing, as it is often called in object-oriented programming, is exactly what it is. You interact with information via methods.

```go
func (*WidgetAlignSystem) Zone(string) *WidgetZone
func (*WidgetAlignSystem) Find(...) int
func (*WidgetAlignSystem) Widtets() []id.Widget

func (*WidgetZone) Section(string) *WidgetSection
func (*WidgetZone) Find(...) int
func (*WidgetZone) Widtets() []id.Widget

func (*WidgetSection) Area(string) *WidgetArea
func (*WidgetSection) Find(...) int
func (*WidgetSection) Widtets() []id.Widget

func (*WidgetArea) Add(...)
func (*WidgetArea) Update(...)
func (*WidgetArea) Remove(...)
func (*WidgetArea) Find(...) int
func (*WidgetArea) Widtets() []id.Widget
```

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+		for s2, a := range s {
+			for _, a2 := range a {
+				if i, h := was.Area(z, s2, a2).Has(wid); h {
+					wloc := WidgetLocation{z, s2, a2}
+					return i, &wloc
+				}
+			}
+		}
+	}
+	return nil, nil
```
rot1024, line 73:
```suggestion
	return 0, nil
```

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+	return &WidgetAlignSystem{}
+}
+
+// Zone will return a specific zone in the align system.
+func (was *WidgetAlignSystem) Zone(zone string) *WidgetZone {
+	if was == nil {
+		return nil
+	}
+	switch zone {
+	case "inner":
```
rot1024, line 87:
```suggestion
	case scene.WidgetZoneInner:
```

constants are not used. Do the same for the others.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+		return &was.inner
+	case "outer":
+		return &was.outer
+	}
+	return nil
```
rot1024, line 92: This branch is not tested. Check other warnings from codecov.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+	return append(array[:index], append([]*id.WidgetID{value}, array[index:]...)...)
+}
+
+// removeInt is used in moveInt to remove the widgetID from original position(index).
+func removeInt(array []*id.WidgetID, index int) []*id.WidgetID {
+	return append(array[:index], array[index+1:]...)
+}
+
+// WidgetAreaFrom will add a slice of widget ids to a specific area of an align system.
+func (was *WidgetAlignSystem) AddAll(wids []*id.WidgetID, align, z, s, a string) {
```
rot1024, line 230:
```suggestion
func (was *WidgetAlignSystem) AddAll(wids []*id.WidgetID, align string, loc *WidgetLocation) {
```

This is exactly where WidgetLocation should be used

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,250 @@
...
+// WidgetSection is the structure of each section of the align system.
+type WidgetSection struct {
+	top    WidgetArea
+	middle WidgetArea
+	bottom WidgetArea
+}
+
+// WidgetArea has the widgets and alignment information found in each part area of a section.
+type WidgetArea struct {
+	widgetIds []*id.WidgetID
```
rot1024, line 29:
```suggestion
	widgetIds []id.WidgetID
```

IDs can only be reference types if you want to make it clear that they can be nil. Most of the time, references are not needed in a ID slice.

Returned value type of methods also need to be fixed

schema.graphql
```diff
@@ -276,6 +308,30 @@ type PluginExtension {
...
+  top: WidgetArea
+  middle: WidgetArea
+  bottom: WidgetArea
+}
+
+type WidgetArea {
+  widgetIds: [ID!]!
+  align: String
+}
+
```
rot1024, line 334: Move them under ScenePlugin and remove `# Widget Align System` because for now comments are for only root entities.

rot1024 (changes requested):
Note the situations where a variable should be a reference type and where it should be a value type. If you treat it as a reference type, it is either when you want to make it clear that it is optional, or when it is a struct and you want to change the field later (i.e., most structs are more convenient to treat as references. (That is, most structs are more convenient to treat as references, but sometimes they are not. Otherwise, use the value type.

internal/adapter/graphql/controller_scene.go
```diff
@@ -56,11 +57,27 @@ func (c *SceneController) AddWidget(ctx context.Context, ginput *AddWidgetInput,
...
+		if l.Index != nil {
+			layout.Index = l.Index
+		}
+		if l.Location != nil {
+			layout.Location = &scene.WidgetLocation{Zone: l.Location.Zone, Section: l.Location.Section, Area: l.Location.Area}
+		}
+		if l.Align != nil {
+			layout.Align = l.Align
+		}
+	}
```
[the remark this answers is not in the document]
KaWaite: I think there are a few issues with the code changes above. But, I got the idea what you were doing and I've updated like so:
```
var layout interfaces.LayoutParams
	if ginput.Layout != nil {
		layout = interfaces.LayoutParams{
			Extended: ginput.Layout.Extended,
			Index: ginput.Layout.Index,
			Location: (*scene.WidgetLocation)(ginput.Layout.Location),
			Align: ginput.Layout.Align,
		}
	}
```
Please let me know if this isn't quite what you were wanting!

pkg/scene/widget_align_system.go
KaWaite: So, this method is necessary after your comment about the Remove and Update methods not needing a Location param, just the WidgetID. Without the Location, we need to search the whole align system for the widgetID. This was the best method I could think of to search the whole tree. 
Also, the index of the found widget is necessary for both the Remove and Update methods(Remove needs to cut out the single widgetID, and Update possibly will be moving the WidgetID's index. I know that could be handled separately in each method, but it would be duplicated code).

I am a little confused with your suggestion though. I understand the message passing, but for each Find method they return an int. What would the Find method be finding? Ultimately, how does that help with searching the whole align system for a widgetID?

pkg/scene/widget_align_system.go
KaWaite: I am using these in the FindWidgetLocation method's for loops' ranges.

pkg/scene/widget_align_system.go
KaWaite: Ok

pkg/scene/widget.go
KaWaite: I am thinking of refactoring the widget, widget system and widget align system into a widget package and creating a builder at that point. If that sounds like a good idea I will leave this as is for now.

pkg/plugin/manifest/convert.go
KaWaite: Sorry, but why is this desirable? Currently, if the Widget doesn't have a DefaultLocation it will be set with `outer, left, top` as a system default. If it has a default location, it is set with that.
So `dl` should never be nil when being set to to `swl.DefaultLocation`.

pkg/plugin/manifest/convert.go
```diff
@@ -129,6 +133,52 @@ func (i Extension) extension(pluginID id.PluginID, sys bool) (*plugin.Extension,
...
+				Section: "left",
+				Area:    "top",
+			},
+		}
+	}
+
+	var swl scene.WidgetLayout
+	swl.Extended = false
+	if l.Extendable == nil {
+		swl.Extendable = true
```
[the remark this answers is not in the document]
KaWaite: I thought that if a plugin dev doesn't specify that, then giving extendability as default would be good.
If you disagree I can change this to false. Either is okay for me!

pkg/plugin/extension.go
```diff
@@ -72,6 +74,11 @@ func (w *Extension) Visualizer() visualizer.Visualizer {
 	return w.visualizer
 }
 
+// WidgetMetaData _
```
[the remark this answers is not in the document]
KaWaite: Removed and removed the other comments as you requested before!

internal/adapter/graphql/controller_scene.go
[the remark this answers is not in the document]
rot1024: Ok

pkg/plugin/manifest/convert.go
rot1024: Hm, I confused because type of DefaultLocation field of WidgetLayout is a reference (`*DefaultLocation`) so it can be nil. If you do not want DefaultLocation to be nil, you can change the type to `DefaultLocation`, because WidgetLocation has no private fields so it can be treated as a simple data container (that is also a "value object").

Note: Strictly speaking, in DDD, the value object should be immutable, but we'll leave that aside for now

pkg/plugin/manifest/convert.go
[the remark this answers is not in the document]
rot1024: ok

pkg/scene/widget_align_system.go
rot1024: First of all, the process of finding a widget in a WidgetAlignSystem is very fast: the CPU is extremely fast, all the data is in memory, and the amount of data is only in the order of 100 at most. You don't need to worry about it.

Since we are assuming that the same widget ID does not exist in more than one place, we don't even need an index number to delete it; we can just give it an ID, search all the areas, and delete it when we find it. The logic can be hidden in the respective Remove methods.

- Remove method of WidgetAlignSystem calls the Remove methods of the two Zones.
- Remove method of Zone calls the Remove methods of the three Sections.
- Remove method of Section calls the three Area Remove methods.
- Remove method of Area loops through the slice with the IDs, and removes the ID when the IDs match.

The key point of this implementation is that WidgetAlignSystem does not manipulate or reference the fields of the Area directly; WidgetAlignSystem has no knowledge of the internals of Area and manipulates them via methods.

Since the number of data is small, this implementation is perfectly fine, and more importantly, it dramatically improves the readability of the code. If you are really concerned about the extra loop, you can make each Remove method return true when the ID is found, and break all loops when it is found.

pkg/plugin/manifest/convert.go
KaWaite: Oh I see what you mean. Then perhaps it is better for all WidgetLayout fields to be mandatory, and just the WidgetLayout to be possibly nil (for non-widget plugins).
I'll update so that default location is necessary.

KaWaite: I have given it more thought. And I think for plugin developers, it would be nice if all Widgetlayout fields are optional except float. Because if float is true, the widget doesn't need any of the other fields.
So I will update WidgetLayout in that way.

pkg/scene/widget_section_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	was := NewWidgetAlignSystem()
+	was.outer.left.top.widgetIds = append(was.outer.left.top.widgetIds, wid)
+	e := was.outer.left.top
+
+	testCases := []struct {
+		Name  string
+		Input struct {
+			id id.WidgetID
+		}
```
rot1024, line 21:
```suggestion
		Input id.WidgetID
```

pkg/scene/widget_system_test.go
```diff
@@ -38,8 +38,9 @@ func TestNewWidgetSystem(t *testing.T) {
 					property:  pr,
 					enabled:   true,
 					widgetLayout: &WidgetLayout{
-						Extendable: &extendable,
-						Extended:   &extended,
+						Extendable: &Extendable{
+							Vertically: &tr},
```
rot1024, line 42:
```suggestion
						Extendable: &Extendable{Vertically: &tr},
```

pkg/scene/widget_zone_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	was := NewWidgetAlignSystem()
+	was.outer.left.top.widgetIds = append(was.outer.left.top.widgetIds, wid)
+	e := was.outer.left.top
+
+	testCases := []struct {
+		Name  string
+		Input struct {
+			id id.WidgetID
+		}
```
rot1024, line 21:
```suggestion
		Input id.WidgetID
```

pkg/scene/widget_zone_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"testing"
+
+	"github.com/reearth/reearth-backend/pkg/id"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWidgetZone_Find(t *testing.T) {
+	wid := id.NewWidgetID()
+
+	was := NewWidgetAlignSystem()
```
rot1024, line 13:
```suggestion
	zone := NewWidgetZone()
```

You cannot use WidgetAlignSystem in `TestWidgetZone_*`

schema.graphql
```diff
@@ -819,6 +880,19 @@ input DeleteProjectInput {
...
+input WidgetLocationInput {
+  zone: String!
+  section: String!
+  area: String!
+}
+
+input WidgetLayoutInput {
+  extended: Boolean
+	location: WidgetLocationInput
+	index: Int
```
rot1024, line 892:
```suggestion
  location: WidgetLocationInput
  index: Int
```

pkg/scene/widget_section_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"testing"
+
+	"github.com/reearth/reearth-backend/pkg/id"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWidgetSection_Find(t *testing.T) {
+	wid := id.NewWidgetID()
+
+	was := NewWidgetAlignSystem()
```
rot1024, line 13:
```suggestion
	section := NewWidgetSection()
```

WidgetAlignSystem cannot be used in `TestWidgetSection_*`

pkg/scene/widget_section.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	s.top.Remove(wid)
+	s.middle.Remove(wid)
+	s.bottom.Remove(wid)
+}
+
+func (s *WidgetSection) Find(wid id.WidgetID) (*int, *WidgetArea) {
+	if s == nil {
+		return nil, nil
```
rot1024, line 24: not tested

pkg/scene/widget_section.go
```diff
@@ -0,0 +1,52 @@
...
+	if wa != nil && i != nil {
+		return i, wa
+	}
+	i2, wa2 := s.middle.Find(wid)
+	if wa2 != nil && i2 != nil {
+		return i2, wa2
+	}
+	i3, wa3 := s.bottom.Find(wid)
+	if wa3 != nil && i3 != nil {
+		return i3, wa3
```
rot1024, line 37: not tested

pkg/scene/widget_section.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (ws *WidgetSection) Area(a string) *WidgetArea {
+	switch a {
+	case WidgetAreaTop:
+		return &ws.top
+	case WidgetAreaMiddle:
+		return &ws.middle
+	case WidgetAreaBottom:
+		return &ws.bottom
```
rot1024, line 49: not tested

pkg/scene/widget_section.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	return nil
```
rot1024, line 51: not tested

pkg/scene/widget_section.go
```diff
@@ -0,0 +1,52 @@
...
+// WidgetSection is the structure of each section of the align system.
+type WidgetSection struct {
+	top    WidgetArea
+	middle WidgetArea
+	bottom WidgetArea
+}
+
+func (s *WidgetSection) Remove(wid id.WidgetID) {
+	if s == nil {
+		return
```
rot1024, line 14: not tested

pkg/scene/widget_area_test.go
```diff
@@ -0,0 +1,143 @@
...
+			tt.Parallel()
+			res := tc.WA.Alignment()
+			assert.Equal(tt, tc.Expected, res)
+		})
+	}
+}
+
+func TestWidgetArea_Has(t *testing.T) {
+	wid := id.NewWidgetID()
+	was := NewWidgetAlignSystem()
```
rot1024, line 107:
```suggestion
	area := NewWidgetArea()
```

WidgetAlignSystem cannot be used in `TestWidgetArea_*`

pkg/scene/widget_area.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Alignment will return the alignment of a specific area.
+func (wa *WidgetArea) Alignment() *string {
+	return &wa.align
+}
+
+func (a *WidgetArea) Remove(wid id.WidgetID) {
+	if a == nil {
+		return
```
rot1024, line 29: not tested

pkg/scene/widget_area.go
```diff
@@ -0,0 +1,64 @@
...
+		if w == wid {
+			a.widgetIds = append(a.widgetIds[:i], a.widgetIds[i+1:]...)
+			return
+		}
+	}
+}
+
+func (a *WidgetArea) Find(wid id.WidgetID) (*int, *WidgetArea) {
+	if a == nil {
+		return nil, nil
```
rot1024, line 42: not tested

pkg/scene/widget_area.go
```diff
@@ -0,0 +1,64 @@
...
+			return &i, a
+		}
+	}
+	return nil, nil
+}
+
+// Has will check a widget area's slice of widgetIds for the specified ID and return a bool value.
+func (wa *WidgetArea) Has(wid id.WidgetID) bool {
+	if wa == nil {
+		return false
```
rot1024, line 56: not tested

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,208 @@
...
+func (was *WidgetAlignSystem) Remove(wid id.WidgetID) {
+	if was == nil {
+		return
+	}
+
+	was.inner.Remove(wid)
+	was.outer.Remove(wid)
+}
+
+func (was *WidgetAlignSystem) FindWidgetLocation(wid id.WidgetID) (*int, *WidgetArea) {
```
rot1024, line 178:
```suggestion
func (was *WidgetAlignSystem) FindWidgetLocation(wid id.WidgetID) (int, *WidgetArea) {
```

The int reference is hard to use, so make it a regular int and return -1 if it is not found.

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,208 @@
...
+	if align != nil {
+		switch *align {
+		case WidgetAlignStart:
+			a.align = WidgetAlignStart
+		case WidgetAlignCenter:
+			a.align = WidgetAlignCenter
+		case WidgetAlignEnd:
+			a.align = WidgetAlignEnd
+		default:
+			a.align = WidgetAlignStart
```
rot1024, line 155: not tested

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	wa := was.Area(loc.Zone, loc.Section, loc.Area)
+	wa.widgetIds = wids
+	wa.align = align
+}
+
+// Update a widget in the align system.
+func (was *WidgetAlignSystem) Update(wid id.WidgetID, l *WidgetLocation, index *int, align *string) {
+	if was == nil {
+		return
```
rot1024, line 138: not tested

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,208 @@
...
+	Area    string
+}
+
+const (
+	WidgetAlignStart  = "start"
+	WidgetAlignCenter = "centered"
+	WidgetAlignEnd    = "end"
+
+	WidgetZoneInner = "inner"
+	WidgetZoneOuter = "outer"
```
rot1024, line 25: This can be moved to `widget_zone.go`

pkg/scene/widget_align_system.go
```diff
@@ -0,0 +1,208 @@
...
+
+	WidgetSectionLeft   = "left"
+	WidgetSectionCenter = "center"
+	WidgetSectionRight  = "right"
```
rot1024, line 29: This can be moved to `widget_section.go`

pkg/scene/widget.go
```diff
@@ -54,6 +72,14 @@ func (w *Widget) Enabled() bool {
...
+func (w *Widget) WidgetLayout() *WidgetLayout {
+	return w.widgetLayout
+}
+
 func (w *Widget) SetEnabled(enabled bool) {
 	w.enabled = enabled
 }
+
+func (w *Widget) SetExtended(extended *bool) {
+	w.widgetLayout.Extended = extended
```
rot1024, line 84: extended should be copied to prevent unexpected changes from outside

pkg/builtin/manifest.yml
```diff
@@ -1212,6 +1212,12 @@ extensions:
     type: widget
     title: Menu
     description: Menu widgets
+    widgetLayout:
+      floating: false
```
rot1024, line 1216:
```suggestion
```

Default can be omitted

pkg/builtin/manifest.yml
```diff
@@ -1386,6 +1394,14 @@ extensions:
     type: widget
     title: Storytelling
     description: SA feature that enables you to create a story. Connect points in a meaningful way so that your information can be consumed and understood easily.
+    widgetLayout:
+      extendable:
+        horizontally: true
+      floating: false
```
rot1024, line 1400:
```suggestion
```

internal/adapter/graphql/convert_alignSystem.go
```diff
@@ -0,0 +1,54 @@
+package graphql
+
+import (
+	"github.com/reearth/reearth-backend/pkg/id"
+	"github.com/reearth/reearth-backend/pkg/scene"
+)
+
+func toAlignSystem(sas *scene.WidgetAlignSystem) *WidgetAlignSystem {
+	widgetAlignDoc := WidgetAlignSystem{
+		Inner: toWidgetZone(sas.Zone("inner")),
```
rot1024, line 10:
```suggestion
		Inner: toWidgetZone(sas.Zone(scene.WidgetZoneInner)),
```

internal/adapter/graphql/convert_alignSystem.go
```diff
@@ -0,0 +1,54 @@
...
+		Outer: toWidgetZone(sas.Zone("outer")),
```
rot1024, line 11:
```suggestion
		Outer: toWidgetZone(sas.Zone(scene.WidgetZoneOuter)),
```

pkg/builtin/manifest.yml
```diff
@@ -1212,6 +1212,12 @@ extensions:
     type: widget
     title: Menu
     description: Menu widgets
+    widgetLayout:
+      floating: false
+      defaultLocation:
+        zone: "outer"
+        section: "left"
+        area: "top"
```
rot1024, line 1220:
```suggestion
        zone: outer
        section: left
        area: top
```

pkg/builtin/manifest.yml
```diff
@@ -1386,6 +1394,14 @@ extensions:
...
     title: Storytelling
     description: SA feature that enables you to create a story. Connect points in a meaningful way so that your information can be consumed and understood easily.
+    widgetLayout:
+      extendable:
+        horizontally: true
+      floating: false
+      defaultLocation:
+        zone: "outer"
+        section: "left"
+        area: "bottom"
```
rot1024, line 1404:
```suggestion
        zone: outer
        section: left
        area: bottom
```

rot1024 (changes requested):

internal/adapter/graphql/convert.go
```diff
@@ -70,11 +70,25 @@ func toWidgetLayout(w *scene.WidgetLayout) *WidgetLayout {
...
+		e = &WidgetExtendable{
+			Vertically:   w.Extendable.Vertically,
+			Horizontally: w.Extendable.Horizontally,
+		}
+		wl = WidgetLayout{
+			Extendable:      e,
+			Extended:        w.Extended,
+			Floating:        w.Floating,
+			DefaultLocation: l,
+		}
```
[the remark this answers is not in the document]
KaWaite: Very strange. The reason I didn't do how you suggest was because I was getting a `nil pointer` type panic.
Now when I try it is working. I wonder why..

pkg/scene/widget.go
KaWaite: Sorry what do you mean exactly by it should be copied? Do you mean the param shouldn't be a pointer? Or do you mean I should set it to a local variable and assign the local variable to w.widgetLayout.Extended? Or something else?

pkg/scene/widget_area_test.go
KaWaite: Other than in testing, I don't see a reason to have NewWidgetArea, NewWidgetSection, NewWidgetZone methods.
Should I create them just to be able to make the tests?

schema.graphql
KaWaite: On my local that extra space isn't showing... weird.

pkg/scene/widget_area_test.go
rot1024: Yes, they should be able to be used independently even if WidgetAlignSystem is not present, so it is better to have their own constructors. Even if you only use them for testing.

pkg/scene/widget.go
rot1024:
```go
e := true
e2 := &e
w.SetExtended(e2)
*e2 = false // This means that you can manipulate the field without going through the
```

KaWaite: Ah, so you are saying I don't need this method? And I should just directly manipulate extended by copying?

rot1024: I mean `extended` needs to be copied in the method:

```go
func (w *Widget) SetExtended(extended *bool) {
	if extended == nil {
		w.widgetLayout.Extended = nil
	} else {
		extended2 := *extended
		w.widgetLayout.Extended = &extended2
	}
```

KaWaite: Sorry, thank you! Time to read up on dereferencing and when/why it is good!

pkg/scene/widget_section_test.go
```diff
@@ -0,0 +1,142 @@
...
+			WS:       ws,
+			Expected: nil,
+		},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.Name, func(tt *testing.T) {
+			tt.Parallel()
+			res := tc.WS.Area(tc.Input)
+			assert.Equal(tt, tc.Expected, res)
```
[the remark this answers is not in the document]
KaWaite: Just to make sure, the shared approach is okay sometimes right? Because quite a few tests need to include one where the method's receiver doesn't exist, which I can't think of how that works if I move the WZ, WA, etc into the for loop.
In other tests (for Widgets, widget system, etc) we don't do as you are suggesting here, but share them and those also have receiver pointers that check if they are nil or not.
Places were data races have occured I removed the `tt.Parallel()` which for obvious reasons avoided the data races. 
Please let me know if you would still like me to do as you suggested here. And if so, please let me know if it is fine I ignore CodeCov's warning that I am not testing the nil receiver.

pkg/scene/widget_area.go
```diff
@@ -0,0 +1,68 @@
...
+type WidgetArea struct {
+	widgetIds []id.WidgetID
+	align     string
+}
+
+var (
+	WidgetAreaTop    = "top"
+	WidgetAreaMiddle = "middle"
+	WidgetAreaBottom = "bottom"
+)
```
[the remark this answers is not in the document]
KaWaite: I tried this but run into issues in plugin/manifest/convert.go when I am converting between packages. Is the above suggested completely necessary? It becomes so that I can't convert values to another library because I can't access a primitive value (string, etc). 

Is there a way to do that? I searched for newType pattern and it looks like it is a Rust thing, and can't find anything on naming constants in Golang in this way. I also looked at some of the packages in the standard library and couldn't find any constants using this pattern.

Please let me know if I can ignore this suggestion or if you know how to convert between packages while using the newtype pattern.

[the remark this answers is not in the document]
rot1024: I mean newtype pattern is just `type WidgetAreaType string`. I just looked them up and found out that they are called "defined types" and "named types".

They can be easily converted into a string:

```go
string(scene.WidgetAreaTop)
```

pkg/scene/widget_section_test.go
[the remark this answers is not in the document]
rot1024: Using the same object in unit tests is not a good idea because it weakens the independence of each test case. Since each test case is less independent, it may not be possible to run them in parallel and data race may occur.

It is easy to make test cases independent. Just confine the creation of structs that may be mutated in each test inside the test case.

Not good:

```go
a := NewHoge()
b := "hoge"

for tc := range testCases {
	tc := tc
	t.Run(tc.Name, func(tt *testing.T) {
		tt.Parallel()

		a.Update() // This may cause data race
        // Using b here is ok because b is used as read-only and not mutated
	})
}
```

Good:

```go
b := "hoge"

for tc := range testCases {
	tc := tc
	t.Run(tc.Name, func(tt *testing.T) {
		tt.Parallel()

		a := NewHoge()
		a.Update() // no problem
	})
}
```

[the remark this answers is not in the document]
rot1024: Unit tests written by others can be left as they are, so for the newly written unit tests in this PR, I would like you to modify them to make the code as good as possible.

pkg/scene/widget_area.go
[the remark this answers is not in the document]
KaWaite: Gotcha, thanks!

rot1024 (approved):

pkg/builtin/manifest.yml
```diff
@@ -1260,6 +1260,11 @@ extensions:
     type: widget
     title: Menu
     description: Menu widgets
+    widgetLayout:
+      defaultLocation:
+        zone: outer
+        section: left
+        area: top
```
rot1024, line 1267: We should keep backward compatibility for already published scenes. Define menu widget as floating and define new widget: Button.

rot1024 (changes requested):

pkg/builtin/manifest.yml
```diff
@@ -1258,8 +1258,14 @@ extensions:
...
     type: widget
-    title: Menu
+    title: Menu (legacy)
     description: Menu widgets
+    widgetLayout:
+      floating: true
+      defaultLocation:
+        zone: outer
+        section: left
+        area: top
```
rot1024, line 1268:
```suggestion
```

Floating widgets have no default location. Is it right?

pkg/builtin/manifest_ja.yml
```diff
@@ -635,6 +635,54 @@ extensions:
             title: リンク
           menuCamera:
             title: カメラ
+  button:
+    title: ボタン
+    description: シーンにボタンを設置し、メニューを表示します。追加したボタンに設定されたアクションタイプによって動作が変わります。\n・リンク：ボタン自体が外部サイトへのリンクになります。\n・メニュー：追加したメニューを開きます\n・カメラアクション：クリック時にカメラを移動します。
```
rot1024, line 640:
```suggestion
    description: |
     　　シーンにボタンを設置し、メニューを表示します。追加したボタンに設定されたアクションタイプによって動作が変わります。
    　　 ・リンク：ボタン自体が外部サイトへのリンクになります。
    　　 ・メニュー：追加したメニューを開きます\n・カメラアクション：クリック時にカメラを移動します。
```

pkg/builtin/manifest_ja.yml
```diff
@@ -579,7 +579,7 @@ extensions:
           item_datanum:
             title: データ(数字)
   menu:
-    title: メニュー
+    title: メニュー (廃止予定)
     description: シーンにボタンを設置し、メニューを表示します。追加したボタンに設定されたアクションタイプによって動作が変わります。\n・リンク：ボタン自体が外部サイトへのリンクになります。\n・メニュー：追加したメニューを開きます\n・カメラアクション：クリック時にカメラを移動します。
```
rot1024, line 583:
```suggestion
    description: |
      シーンにボタンを設置し、メニューを表示します。追加したボタンに設定されたアクションタイプによって動作が変わります。
      ・リンク：ボタン自体が外部サイトへのリンクになります。
      ・メニュー：追加したメニューを開きます\n・カメラアクション：クリック時にカメラを移動します。
```

rot1024 (approved):

pkg/builtin/manifest.yml
KaWaite: Ah that's true! Thank you

rot1024 (approved):

PR #23, opened 2021-07-13
fix: import dataset from google sheet bug

yk-eukarya:
# Overview

## What I've done
Check if the column has an empty label => ignore it

## Note
This will affect the import from file also in the same way


pkg/dataset/csvparser.go
```diff
@@ -87,7 +88,7 @@ func (p *DatasetCSVParser) GuessSchema(sid id.SceneID) error {
 		if h == "lng" {
 			haslng = true
 		}
-		if h != "lng" && h != "lat" {
+		if h != "lng" && h != "lat" && strings.TrimSpace(h) != "" {
```
rot1024, line 91: Can you explain how this prevents bugs?

rot1024 (approved):

HideBa (approved):

PR #25, opened 2021-07-16
fix: KML coordinates splitting 

mimoham24:
# Overview
Fix the bug that KML coordinates were not decoded for "\t" separator
## What I've done
replaced strings.Split() with strings.Fields() to cover all white spaces

## How I tested
 separated coordinates for each white-space


rot1024 (approved):
I've never heard of `strings.Fields` before. That's a useful function.

PR #26, opened 2021-07-19
feat: Refine dataset import

yk-eukarya:
# Overview
Add ability to set the layer name from dataset column

## What I've done
- Extend AddLayerGroupInput to accept optional markerTitleFieldId
- Use selected Field value as layer name

specially the graphQL schema so we can start the front-end development

## Memo
The layer primitive type was implemented I've just tested it

### DIFF
--- a/internal/usecase/interactor/layer.go
+++ b/internal/usecase/interactor/layer.go
@@ -354,6 +354,7 @@ func (i *Layer) AddGroup(ctx context.Context, inp interfaces.AddLayerGroupInput,
 	}
 
 	// create item layers
+	layerTitleFieldId := inp.MarkerTitleFieldId
 	representativeFieldID := datasetSchema.RepresentativeFieldID()
 	layerItems := make([]*layer.Item, 0, len(ds))
 	layerItemProperties := make([]*property.Property, 0, len(ds))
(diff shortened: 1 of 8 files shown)


internal/usecase/interactor/layer.go
```diff
@@ -365,6 +366,11 @@ func (i *Layer) AddGroup(ctx context.Context, inp interfaces.AddLayerGroupInput,
 		if rf := ds.FieldRef(representativeFieldID); rf != nil && rf.Type() == dataset.ValueTypeString {
 			name = rf.Value().Value().(string)
 		}
+		if name == "" && layerTitleFieldId != nil {
+			if rf := ds.FieldRef(layerTitleFieldId); rf != nil && rf.Type() == dataset.ValueTypeString {
+				name = rf.Value().Value().(string)
+			}
+		}
```
rot1024, line 373:
```suggestion
```

representativeFieldID is already available for such usecase, so just set layerTitleFieldId as representativeFieldID instead of setting layer name manually.

Pass the representativeFieldID to the logic that generates the dataset schema and set the representativeFieldID internally.

schema.graphql
```diff
@@ -995,6 +996,7 @@ input AddLayerGroupInput {
   index: Int
   linkedDatasetSchemaID: ID
   name: String
+  markerTitleFieldId: DatasetSchemaFieldID
```
rot1024, line 999: The layer to be added is not necessarily a marker, so this name is too specific; how about representativeFieldId?

rot1024 (changes requested):

schema.graphql
yk-eukarya: Resolved

internal/usecase/interactor/layer.go
yk-eukarya: We can not set the representativeFieldID in the dataset schema, because according to the feature design the user should be able to set the representativeFieldID every time he adds the dataset as layer group

rot1024: Oh, I see. I was mistaken.

rot1024 (approved):

PR #36, opened 2021-08-04
fix: dataset layers are not exported correctly

mimoham24:
# Overview
fixing the layer exporting bug
## What I've done
encoder now fetches the property values first, if no values found -> fetch values from dataset

## How I tested
export all types "KML, CZML, GeoJSON, SHP"


pkg/layer/encoding/czml.go
```diff
@@ -56,13 +56,26 @@ func (e *CZMLEncoder) encodeSingleLayer(li *merging.SealedLayerItem) (*czml.Feat
 		if li.Property.Field("location") != nil {
```
rot1024, line 56:
```suggestion
		if f := li.Property.Field("location"); f != nil {
```

More readable code

Change other parts of the file in the same way.

pkg/layer/encoding/czml.go
```diff
@@ -56,13 +56,26 @@ func (e *CZMLEncoder) encodeSingleLayer(li *merging.SealedLayerItem) (*czml.Feat
...
 			latlng, ok = li.Property.Field("location").PropertyValue.ValueLatLng()
```
rot1024, line 57:
```suggestion
			latlng, ok = f.PropertyValue.ValueLatLng()
```

pkg/layer/encoding/czml.go
```diff
@@ -56,13 +56,26 @@ func (e *CZMLEncoder) encodeSingleLayer(li *merging.SealedLayerItem) (*czml.Feat
...
 			if !ok {
-				return nil, errors.New("invalid value type")
+				dsll := li.Property.Field("location").DatasetValue.ValueLatLng()
```
rot1024, line 59:
```suggestion
				dsll := f.DatasetValue.ValueLatLng()
```

rot1024 (dismissed):
Changes are ok but I've tried to make the code a little more readable.

HideBa (approved):
Thanks! 👍

HideBa (approved):

rot1024 (approved):

PR #41, opened 2021-08-31
fix: layer importing bug

mimoham24:
# Overview
https://github.com/reearth/reearth/issues/109
## What I've done
- fix stylelement assertion for nil values
- return an appropriate error for unsupported geojson geometric types
## How I tested
- importing geojson content without any style
## Which point do I want you to review particularly
to test the provided data sample(https://gist.github.com/wavded/1200773?short_path=99c1af9)


rot1024 (approved):
The implementation of layer import and export, for example, the createProperty function is getting too big, so it would be great to split it into smaller functions.

PR #42, opened 2021-09-01
fix: skip nil geometries

mimoham24:
# Overview
skip a geoJSON feature when it doesn't has a geometry to decode
## What I've done
skip nil geos 

## Which point do I want you to review particularly
validation function


rot1024 (approved):

HideBa (approved):
Thank you!

PR #44, opened 2021-09-02
feat: tag system data-layer (mongo)

mimoham24:
# Overview
tag system repos, docs, and models
## What I've done
1- create tag system datalayer interface
2- implement the interface
3- create MongoDB persistent layer for tag system 
## What I haven't done
implement data persistent for memory

### DIFF
--- a/pkg/tag/group_builder.go
+++ b/pkg/tag/group_builder.go
@@ -16,6 +16,14 @@ func NewGroup() *GroupBuilder {
 	return &GroupBuilder{g: &Group{}}
 }
 
+func GroupFromTag(t Tag) *Group {
+	li, ok := t.(*Group)
+	if !ok {
+		return nil
+	}
+	return li
+}
+
 func (b *GroupBuilder) Build() (*Group, error) {
 	if id.ID(b.g.id).IsNil() {
 		return nil, id.ErrInvalidID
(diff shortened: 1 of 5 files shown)


pkg/tag/group_builder.go
```diff
@@ -16,6 +16,14 @@ func NewGroup() *GroupBuilder {
 	return &GroupBuilder{g: &Group{}}
 }
 
+func GroupFromTag(t Tag) *Group {
```
rot1024, line 19:
```suggestion
func GroupFrom(t Tag) *Group {
```

Since it is clear from the package name and type signature that it accepts tags, we can make the function name shorter

- `tag.GroupFromTag`: "tag" is repeated twice
- `tag.GroupFrom`: good

pkg/tag/item_builder.go
```diff
@@ -10,6 +10,14 @@ func NewItem() *ItemBuilder {
 	return &ItemBuilder{i: &Item{}}
 }
 
+func ItemFromTag(t Tag) *Item {
```
rot1024, line 13:
```suggestion
func ItemFrom(t Tag) *Item {
```

rot1024 (approved):
Good, but It would be nice to have unit tests for TagDocument on mongodoc package!

HideBa (approved):
LGTM

PR #45, opened 2021-09-10
feat: create tag group and tag item

mimoham24:
# Overview
`createTagItem` and `createTagGroup` mutations 
## What I've done
- create graphQL schema
- create the presentation layer
- create dummy infrastructure/memory layer (will create another PR implementing it)
- create use-case layer
- implement creating tag items and groups mutations

## What I haven't done
implementing `infrastructure/memory`
## How I tested
providing dummy data to the mutations


HideBa (approved):
LGTM

internal/usecase/interactor/tag.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err2 := tx.End(ctx); err == nil && err2 != nil {
+			err = err2
+		}
+	}()
+
+	if err := i.CanReadScene(ctx, inp.SceneID, operator); err != nil {
```
rot1024, line 39:
```suggestion
	if err := i.CanWriteScene(ctx, inp.SceneID, operator); err != nil {
```

internal/usecase/interactor/tag.go
```diff
@@ -0,0 +1,99 @@
...
```
rot1024, line 77:
```suggestion
	if err := i.CanWriteScene(ctx, inp.SceneID, operator); err != nil {
```

schema.graphql
```diff
@@ -706,6 +706,33 @@ type MergedInfoboxField {
...
+  linkedDatasetSchema: DatasetSchema @goField(forceResolver: true)
+  linkedDataset: Dataset @goField(forceResolver: true)
+  linkedDatasetField: DatasetField @goField(forceResolver: true)
+}
+
+type TagGroup implements Tag {
+  id: ID!
+  sceneId: ID!
+  label: String!
+  tags: [ID]
```
rot1024, line 731:
```suggestion
  tags: [ID!]!
```

schema.graphql
```diff
@@ -1046,6 +1073,20 @@ input AddDatasetSchemaInput {
...
+  label: String!
+  linkedDatasetSchemaID: ID
+  linkedDatasetID: ID
+  linkedDatasetField: ID
+}
+
+input CreateTagGroupInput{
+  sceneId: ID!
+  label: String!
+  tags: [ID]
```
rot1024, line 1087:
```suggestion
  tags: [ID!]!
```

schema.graphql
```diff
@@ -1046,6 +1073,20 @@ input AddDatasetSchemaInput {
...
+  sceneId: ID!
+  label: String!
+  linkedDatasetSchemaID: ID
+  linkedDatasetID: ID
+  linkedDatasetField: ID
+}
+
+input CreateTagGroupInput{
+  sceneId: ID!
+  label: String!
```
rot1024, line 1086:
```suggestion
  label: String
```

When the user does not want to update the name, the user won't set the label field

rot1024 (changes requested):

schema.graphql
rot1024: Note: apply this suggestion only for UpdateTag*Input

rot1024 (approved):

PR #49, opened 2021-09-15
Tag/update group label

mimoham24:
# Overview
update tag group label
## What I've done
I've created a mutation to update tag group label
## How I tested
dummy data


schema.graphql
```diff
@@ -1486,6 +1496,7 @@ type Mutation {
   # Tag
   createTagItem(input: CreateTagItemInput!): CreateTagItemPayload
   createTagGroup(input: CreateTagGroupInput!): CreateTagGroupPayload
+  renameTagGroup(input: RenameTagGroupInput!): RenameTagGroupPayload
```
rot1024, line 1499:
```suggestion
  updateTag(input: UpdateTagInput!): UpdateTagPayload
```

"updateTag" is better because:

- We can omit work to manage 2 similar mutations (renameTagGroup and renameTagIItem)
- For now, there are not different between tag items and tag groups for fields users want to update (label only)
- If a user-updatable tag field is added to tags in the future, this mutation will no longer just rename tags

Rename usecase layer also.

rot1024 (changes requested):

HideBa (approved):

PR #50, opened 2021-09-15
feat: attach/detach tag from layer

mimoham24:
# Overview

## What I've done
created two mutations to attach and detach tag from layer

## What I haven't done
test units
## How I tested
dummy data


schema.graphql
```diff
@@ -1289,6 +1302,14 @@ type CreateTagGroupPayload{
   tag: TagGroup!
 }
 
+type AttachTagToLayerPayload{
+  layer: Layer!
+}
+
+type DetachTagFromLayerPayload{
+    layer: Layer!
```
rot1024, line 1310: indent is difference from AttachTagToLayerPayload
```suggestion
  layer: Layer!
```

schema.graphql
```diff
@@ -618,6 +619,7 @@ type LayerItem implements Layer {
   merged: MergedLayer @goField(forceResolver: true)
   scene: Scene @goField(forceResolver: true)
   scenePlugin: ScenePlugin @goField(forceResolver: true)
+  tags: [ID]
```
rot1024, line 622:
```suggestion
  tags: [ID!]
```

schema.graphql
```diff
@@ -642,6 +644,7 @@ type LayerGroup implements Layer {
   layers: [Layer]! @goField(forceResolver: true)
   scene: Scene @goField(forceResolver: true)
   scenePlugin: ScenePlugin @goField(forceResolver: true)
+  tags: [ID]
```
rot1024, line 647:
```suggestion
  tags: [ID!]
```

`[ID!]` allows [null, null, ID] but `[ID!]` does not allow it

schema.graphql
```diff
@@ -586,6 +586,7 @@ interface Layer {
   plugin: Plugin
   extension: PluginExtension
   scenePlugin: ScenePlugin
+  tags: [ID]
```
rot1024, line 589:
```suggestion
  tags: [ID!]
```

pkg/layer/layer.go
```diff
@@ -4,10 +4,17 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/reearth/reearth-backend/pkg/tag"
+
```
rot1024, line 8:
```suggestion
```

pkg/layer/group_builder.go
```diff
@@ -110,3 +111,8 @@ func (b *GroupBuilder) LinkedDatasetSchema(linkedDatasetSchema *id.DatasetSchema
 	b.l.linkedDatasetSchema = linkedDatasetSchema.CopyRef()
 	return b
 }
+
+func (b *GroupBuilder) Tags(tags tag.List) *GroupBuilder {
```
rot1024, line 115: Write unit tests

pkg/layer/builder.go
```diff
@@ -71,3 +72,8 @@ func (b *Builder) Infobox(infobox *Infobox) *Builder {
 	b.base.infobox = infobox
 	return b
 }
+
+func (b *Builder) Tags(tags tag.List) *Builder {
+	b.base.tags = tags
+	return b
```
rot1024, line 78: NewTagList returns `*tag.List` but this func accepsts `tag.List`, which does not make sense.

```suggestion
func (b *Builder) Tags(tags *tag.List) *Builder {
	b.base.tags = tags.Clone()
	return b
```

Define Copy method to Tag

pkg/layer/layer.go
```diff
@@ -72,6 +82,7 @@ type layerBase struct {
 	property  *id.PropertyID
 	infobox   *Infobox
 	scene     id.SceneID
+	tags      tag.List
```
rot1024, line 85:
```suggestion
	tags      *tag.List
```

rot1024 (changes requested):
- Pay attention to the receiver type of the struct's method, and decide whether to treat it as a reference type or a value type. If you treat it as a value type, it means that the method cannot modify itself. Make sure you understand this difference.
- Add unit tests for layer package

pkg/layer/group_builder_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/reearth/reearth-backend/pkg/id"
+	"github.com/reearth/reearth-backend/pkg/tag"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGroupBuilder_Tags(t *testing.T) {
+	tags := []id.TagID{id.NewTagID()}
+	b := NewGroup().NewID().Tags(tag.NewListFromTags(tags)).MustBuild()
+	assert.Equal(t, tags, b.Tags().Tags())
```
rot1024, line 14:
```suggestion
	b := NewGroup().NewID().Tags(tag.NewListFromTags(tags))
	assert.Same(t, tags, b.l.tags)
```

Let's think a little more carefully about whether we should use Equal or Same for assertions. If the implementation is just an assignment, then Same should be used to assert that the value being assigned is exactly the same as the expected value. On the other hand, if there is a possibility that the value will be copied, then Equal is fine. If you use Equal for everything, unit tests will not be able to detect if a value is unintentionally copied when testing an implementation of a function that does not copy values.

Also, accessing private fields of TagGroupBuilder and TagGroup is ok because they are just the unit test target.

pkg/layer/group_test.go
```diff
@@ -135,3 +139,15 @@ func TestGroup_MoveLayerFrom(t *testing.T) {
...
+	tt := id.NewTagID()
+	err := group.AttachTag(tt)
+	assert.NoError(t, err)
+	tl := tags
+	tl = append(tl, tt)
+	assert.Equal(t, tl, group.Tags().Tags())
+	err = group.DetachTag(tt)
+	assert.NoError(t, err)
+	assert.Equal(t, tags, group.Tags().Tags())
+}
```
rot1024, line 153: I found that this test file is not very good, because multiple functions are referencing and mutating a single variable. In general, in many cases, code that mutates global variables (or their equivalents, in this case variables defined with var) is an anti-pattern to avoid. It is true that Go's unit tests are executed sequentially within the same package and no problem to run the tests, but the independence of functions is compromised and the implementation becomes difficult to understand.

But this time, let's leave it at that. We'll have a chance to refactor it later.

rot1024 (changes requested):

pkg/layer/group_builder_test.go
mimoham24: @rot1024 I believe equal is good here because tags are passed by value (please see `NewListFromTags(tags []id.TagID)`) but anyway I changed the test unit logic

rot1024: Yes, right. `Same` doesn't work for values (not references). But we have two choices to  assert a reference: Same or Equal.

rot1024 (approved):

PR #51, opened 2021-09-15
chore: update golang version and modules

mimoham24:
# Overview
7upgrading to glang version 1.15
## What I've done
- update golang locally
- `go get -u`
- `go mod tidy -go=1.17` (equals to `go mod edit -go=1.17` and `go mod tidy`)
- update go version for `github workflows` and `DOCKER file`
## How I tested
- run test
- run and build the project
## Memo
language changes:
- Conversions from slice to array pointer: An expression s of type []T may now be converted to array pointer type *[N]T. If a is the result of such a conversion, then corresponding indices that are in range refer to the same underlying elements: &a[i] == &s[i] for 0 <= i < N. The conversion panics if len(s) is less than N.
- unsafe.Add: unsafe.Add(ptr, len) adds len to ptr and returns the updated pointer unsafe.Pointer(uintptr(ptr) + uintptr(len)).
- unsafe.Slice: For expression ptr of type *T, unsafe.Slice(ptr, len) returns a slice of type []T whose underlying array starts at ptr and whose length and capacity are len.


rot1024 (changes requested):
Change `.github/workflows/*.yml` and `Dockerfile` also

rot1024 (approved):

PR #54, opened 2021-09-24
feat: tag system queries

mimoham24:
# Overview
tag system queries 
## What I've done
- Query to fetch the tags under the scene
- Query to fetch the tags attached on the layer
- Query to fetch the tag groups under the scene
- Query to fetch the tag groups attached on the layer


HideBa (approved):
LGTM 👍

PR #55, opened 2021-09-28
fix: invalid extension data from GraphQL, plugin manifest schema improvement, more friendly error from manifest parser

rot1024:
- Fix invalid extension title and description are returned on GraphQL
- Make errors more friendly from plugin manifest parser
- Plugin manifest schema improvement
  - Make "visualizer" field optional in extensions
  - Rename title field to "name" in root and extensions


rot1024:
The commit from this PR is deleted because force push to main branch

PR #58, opened 2021-09-28
Tag/remove tag

mimoham24:
# Overview
remove tag rules:
- can't remove a tag non-empty tag groups
- a removed tag group need to be detached from the layer
- a removed tag item should be detached from both tag group and layer
## What I've done
creat:
- use-cases
- gql mutaions
- datalayers
## How I tested
dummy data


HideBa (approved):
LGTM 👍

PR #59, opened 2021-10-06
chore: use golangci-lint-action

rot1024:
Introduce [golangci-lint-action](https://github.com/golangci/golangci-lint-action) to reduce the execution time of actions


rot1024:
golangci-lint-action doesn't seem to work...

PR #65, opened 2021-10-17
refactor: signup

mimoham24:
# Overview


pkg/user/user.go
```diff
@@ -118,3 +120,12 @@ func (u *User) RemoveAuthByProvider(provider string) bool {
 func (u *User) ClearAuths() {
 	u.auths = []Auth{}
 }
+
+func EncodePassword(pass string) ([]byte, error) {
```
rot1024, line 124: The password encoding process should be completely confined within the domain model and completely invisible from the outside. The use case layer should not know how the password is encoded at all. This is a security issue and should be implemented carefully.

`EncodePassoword` func should be private and used only by `func (*User) SetPassword(password string)` setter method. Also `comparePassword` function is needed and should be called by `func (*User) ComparePassword(password string) bool` func. 

```suggestion
func comparePassword(pass, pass2 string) bool {
	// ...
}

func cncodePassword(pass string) ([]byte, error) {
```

rot1024 (changes requested):

pkg/user/auth.go
```diff
@@ -18,3 +21,17 @@ func AuthFromAuth0Sub(sub string) Auth {
...
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+	b := make([]rune, length)
+	for i := range b {
+		b[i] = letters[rand.Intn(len(letters))]
+	}
+
+	return Auth{
+		Provider: "reearth",
+		Sub:      string(b),
```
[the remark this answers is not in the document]
mimoham24: @yk-eukarya FYI, sub is the same as userID

pkg/user/user.go
```diff
@@ -121,11 +127,38 @@ func (u *User) ClearAuths() {
...
+	p, err := encodePassword(pass)
+	if err != nil {
+		return err
+	}
+	u.password = p
+	return nil
+}
+
+func (u *User) MatchPassword(pass string) (bool, error) {
+	return verifyPassword(pass, u.password)
```
rot1024, line 140: write unit test for this method

pkg/user/user.go
```diff
@@ -121,11 +127,38 @@ func (u *User) ClearAuths() {
...
 }
 
-func EncodePassword(pass string) ([]byte, error) {
+func (u *User) SetPassword(pass string) error {
+	p, err := encodePassword(pass)
+	if err != nil {
+		return err
+	}
+	u.password = p
+	return nil
```
rot1024, line 136: write unit test for this method

pkg/user/user.go
```diff
@@ -1,11 +1,17 @@
...
 import (
+	"errors"
+
 	"github.com/matthewhartstonge/argon2"
 	"github.com/reearth/reearth-backend/pkg/id"
 	"golang.org/x/text/language"
 )
 
+var (
+	ErrEncodingPass = errors.New("error encoding password")
```
rot1024, line 12:
```suggestion
	ErrEncodingPassword = errors.New("error encoding password")
```

rot1024 (changes requested):

pkg/user/user_test.go
```diff
@@ -276,3 +276,73 @@ func TestUser_UpdateName(t *testing.T) {
...
+			},
+			want: "test",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(tt *testing.T) {
+			u := &User{}
+			_ = u.SetPassword(tc.args.pass)
+			got, _ := verifyPassword(tc.want, u.password)
+			assert.True(tt, got)
```
rot1024, line 345: Assert the error. If verifyPassword has been broken, this test should detect it. It's important for security.

```suggestion
			got, err := verifyPassword(tc.want, u.password)
			assert.NoError(tt, err)
			assert.True(tt, got)
```

pkg/user/user_test.go
```diff
@@ -276,3 +276,73 @@ func TestUser_UpdateName(t *testing.T) {
...
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(tt *testing.T) {
+			u := &User{
+				password: tc.password,
+			}
+			got, err := u.MatchPassword(tc.args.pass)
+			assert.Equal(tt, tc.want, got)
+			assert.Equal(tt, tc.wantErr, err != nil)
```
rot1024, line 318: I recommend you to read https://pkg.go.dev/github.com/stretchr/testify/assert once. Many assertion funcs are available.

```suggestion
			if tc.wantErr {
				assert.Error(tt, err)
			} else {
				assert.NoError(tt, err)
			}
```

rot1024 (changes requested):

rot1024 (commented):
Also warning from golangci-lint should be fixed:

```
internal/adapter/http/user.go:49:2: SA9003: empty branch (staticcheck)
	if input.Email != nil {
	^
```

pkg/user/builder.go
```diff
@@ -48,6 +55,15 @@ func (b *Builder) Email(email string) *Builder {
...
 
+func (b *Builder) Password(strPassword string, binPassword []byte) *Builder {
+	if binPassword == nil {
+		b.p = &strPassword
+	} else {
+		b.u.password = binPassword
+	}
+	return b
+}
+
```
rot1024, line 66: String password is not used. And slice should be copied before setting them to flelds

Note: `append(a[:0:0], a...)` is copying a slice

```suggestion
func (b *Builder) Password(password []byte) *Builder {
	if len(password) == 0 {
		b.u.password = nil
	} else {
		b.u.password = append(p[:0:0], p...)
	}
	return b
}

```

pkg/user/initializer.go
```diff
@@ -8,7 +8,8 @@ import (
 type InitParams struct {
 	Email    string
 	Name     string
-	Auth0Sub string
+	Sub      *Auth
+	Password string
```
rot1024, line 12:
```suggestion
	Password []byte
```

pkg/user/user.go
```diff
@@ -39,6 +47,10 @@ func (u *User) Theme() Theme {
 	return u.theme
 }
 
+func (u *User) Password() []byte {
+	return u.password
+}
+
```
rot1024, line 53:
```suggestion
func (u *User) PasswordExists() bool {
	return u != nil && len(u.password) > 0
}
```

Password getter is unnecessary and weaken security. SetPassword and MatchPassword are enough. Instead, PasswordExists method can be useful in the future.

rot1024 (changes requested):

pkg/user/user.go
```diff
@@ -130,3 +142,39 @@ func (u *User) RemoveAuthByProvider(provider string) bool {
...
+	p, err := encodePassword(pass)
+	if err != nil {
+		return err
+	}
+	u.password = p
+	return nil
+}
+
+func (u *User) MatchPassword(pass string) (bool, error) {
+	return verifyPassword(pass, u.password)
```
rot1024, line 156: Add nil check

```suggestion
	if u == nil || len(u.password) == 0 {
		return false, nil
	}
	return verifyPassword(pass, u.password)
```

rot1024 (changes requested):

pkg/user/builder.go
mimoham24: @rot1024 string password is used for creating a new password that is not created yet .... in other words we need to pass an uncoded password and code it during the building process

bin pass is for the infrastructure layer where I need to fetch the stored binary password and pass it to the user model

pkg/user/user.go
mimoham24: @rot1024 we need a getter to store the password to the user doc

rot1024: ok

rot1024 (changes requested):

pkg/user/user.go
```diff
@@ -1,19 +1,29 @@
 package user
 
 import (
+	"errors"
+
+	"github.com/matthewhartstonge/argon2"
```
rot1024, line 6: - `golang.org/x/crypto/argon2` is better than `github.com/matthewhartstonge/argon2` because it's strongly recommended to use the official implementation for security.
- Use `golang.org/x/crypto/bcrypt`: I had said we would use Argon2, but I'm afraid I have to retract that: Argon2 requires various parameters to be encoded and they have to be stored with the hash. Bcrypt, on the other hand, requires no special parameters to encode and is very easy to use.

rot1024 (changes requested):

rot1024 (approved):

PR #68, opened 2021-11-02
feat: support opentelemetry

int-tt:
# Overview
Support for opentelemetry and net/pprof.

## What I've done
Added support for open telemetry and sending metrics via GCP CloudTrace.

## How I tested
Add the fllowing contents to the .env file.
```
REEARTH_TRACER="gcp"
REEARTH_TRACERSAMPLE="10.0"
GOOGLE_CLOUD_PROJECT=example-project
GOOGLE_APPLICATION_CREDENTIALS=./sa.json
```
Launch the app and process the request. then I confirmed that the data was reflected in the cloud trace.


rot1024 (approved):
Thank you!

PR #72, opened 2021-11-09
feat: authentication system server

yk-eukarya:
closes reearth/reearth#135 & reearth/reearth#142


internal/app/oauth/storage.go
```diff
@@ -0,0 +1,337 @@
...
+			responseTypes:   []oidc.ResponseType{oidc.ResponseTypeCode},
+			grantTypes:      []oidc.GrantType{oidc.GrantTypeCode, oidc.GrantTypeRefreshToken},
+			redirectURIs:    []string{"http://localhost:3000"},
+			allowedScopes:   []string{"openid", "profile", "email"},
+			loginURI:        "http://localhost:3000/login?id=%s",
+			iDTokenLifetime: 5 * time.Minute,
+			clockSkew:       0,
+			devMode:         true,
+		},
+	}
```
rot1024, line 65:
```suggestion
	client := initClient(dev)
	clients = map[string]op.Client{
	   client.GetID(): client,
	}
```

Separate this go file to define the shared client.

```go
func initClient(dev bool) op.Client {
	return &ConfClient{
		// ...
		devMode: dev,
	}
}
```

Also consider:

- `applicationType `: Have you considered `ApplicationTypeWeb`?
- `authMethod`: Make sure that `AuthMethodNone` is ok. I think it means the client has no client secret.

internal/app/oauth/request.go
```diff
@@ -0,0 +1,116 @@
...
+	now := time.Now()
+	a.authorizedAt = &now
+}
+
+func unique(list []string) []string {
+	allKeys := make(map[string]bool)
+	var uniqueList []string
+	for _, item := range list {
+		if _, value := allKeys[item]; !value {
+			allKeys[item] = true
```
rot1024, line 111:
```suggestion
	allKeys := make(map[string]struct{})
	var uniqueList []string
	for _, item := range list {
		if _, ok := allKeys[item]; !ok {
			allKeys[item] = struct{}{}
```

`struct{}` is the simplest data type that consume 0 byte in memory

internal/app/oauth/storage.go
```diff
@@ -0,0 +1,337 @@
...
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"errors"
+	"math/big"
+	mRand "math/rand"
```
rot1024, line 11:
```suggestion
	mrand "math/rand"
```

Uppercase is not used in package name

internal/app/auth-srv.go
```diff
@@ -0,0 +1,207 @@
...
+	http1 "github.com/reearth/reearth-backend/internal/adapter/http"
+	"github.com/reearth/reearth-backend/internal/app/oauth"
+	"github.com/reearth/reearth-backend/internal/usecase/interactor"
+
+	"github.com/golang/gddo/httputil/header"
+)
+
+func AuthEndPoints(e *echo.Echo, r *echo.Group, cfg *ServerConfig) {
+
+	usersController := http1.NewUserController(interactor.NewUser(cfg.Repos, cfg.Gateways, cfg.Config.SignupSecret))
```
rot1024, line 24: This handler behaves like a special controller, so you can use usecase directly instead of using a controller.
```suggestion
	userUsecase :=interactor.NewUser(cfg.Repos, cfg.Gateways, cfg.Config.SignupSecret)
```

internal/app/auth-srv.go
```diff
@@ -0,0 +1,207 @@
...
+
+	ctx := context.Background()
```
rot1024, line 26: Receive ctx from outside: `func AuthEndPoints(ctx context.Context)`

internal/app/oauth/storage.go
```diff
@@ -0,0 +1,337 @@
...
+	clients   map[string]op.Client
+	requests  map[string]AuthRequest
+	keySet    jose.JSONWebKeySet
+)
+
+type AuthStorage struct {
+	key *rsa.PrivateKey
+}
+
+type AuthSrvConfig struct {
```
rot1024, line 32:
```suggestion
type StorageConfig struct {
```

Simpler and understandable name

internal/app/oauth/storage.go
```diff
@@ -0,0 +1,337 @@
...
+)
+
+var (
+	appConfig *AuthSrvConfig
+	clients   map[string]op.Client
+	requests  map[string]AuthRequest
+	keySet    jose.JSONWebKeySet
+)
+
+type AuthStorage struct {
```
rot1024, line 28:
```suggestion
type Storage struct {
```

package name is already `oauth` (my suggestion is `appauth`) so `Auth` prefix is not needed.

You can see this type as following:

```go
oauth.AuthStorage{...} // not good
oauth.Storage{...} // good: ok, it's a storage of oauth
```

internal/app/oauth/client.go
```diff
@@ -0,0 +1,98 @@
...
+func (c *ConfClient) ResponseTypes() []oidc.ResponseType {
+	return c.responseTypes
+}
+
+func (c *ConfClient) GrantTypes() []oidc.GrantType {
+	return c.grantTypes
+}
+
+func (c *ConfClient) DevMode() bool {
+	return c.devMode && appConfig.Debug
```
rot1024, line 68:
```suggestion
	return c.devMode
```

Incomprehensible code. 

Until I read this line, I wouldn't expect it to reflect the application settings. To make the intention clearer, set `appCofnig.Debug` to `devMode` only during initialization.

internal/app/oauth/request.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (a *AuthRequest) GetAMR() []string {
+	return []string{
+		"password",
+	}
+}
+
+func (a *AuthRequest) GetAudience() []string {
+	audiences := []string{
+		appConfig.Domain,
```
rot1024, line 42:
```suggestion
		a.Domain,
```

AuthRequest can have Domain instead of referring a global variable.

internal/app/oauth/request.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	if appConfig.Debug {
```
rot1024, line 45:
```suggestion
	if a.Debug {
```

AuthRequest can have Debug instead of referring a global variable.

internal/app/oauth/request.go
```diff
@@ -0,0 +1,116 @@
...
+	subject       string
+	code          string
+	state         string
+	ResponseType  oidc.ResponseType
+	scopes        []string
+	RedirectURI   string
+	Nonce         string
+	CodeChallenge *oidc.CodeChallenge
+	createdAt     time.Time
+	authorizedAt  *time.Time
```
rot1024, line 23:
```suggestion
	authorizedAt  *time.Time
	Domain        string
	Debug         bool
```

Inject appConfig's fields to these fields during initialization of AuthRequest

internal/app/oauth/storage.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/oklog/ulid"
+	"gopkg.in/square/go-jose.v2"
+)
+
+var (
+	appConfig *AuthSrvConfig
+	clients   map[string]op.Client
+	requests  map[string]AuthRequest
+	keySet    jose.JSONWebKeySet
+)
```
rot1024, line 26:
```suggestion
```

This code works, but it is very hard to understand the code because some functions modify global variables. 

Remember that **MODIFYING GLOBAL VARIABLES IS EVIL** in 99% cases. 

In golang, the special case is modifying them only in `func init()` (automatically called when the app is initialized).

Better code is like this:

```go
type AuthStorage struct { // note: I suggest renaming AuthStorage Storage
	clients   map[string]op.Client // came from the global variables
	requests  map[string]AuthRequest // came from the global variables
	keySet    jose.JSONWebKeySet // came from the global variables
    domain    string // came from authSrvConfig
    debug     bool // came from authSrvConfig
	key       *rsa.PrivateKey
}
```

rot1024 (changes requested):

internal/app/oauth/storage.go
```diff
@@ -0,0 +1,337 @@
...
+	requests[request.ID] = *request
+	return request, nil
+}
+
+func (s *AuthStorage) AuthRequestByID(_ context.Context, requestID string) (op.AuthRequest, error) {
+
+	if requestID == "" {
+		return nil, errors.New("invalid id")
+	}
+	request, exists := requests[requestID]
```
rot1024, line 153: Ensure this code is called by goroutine-safe func. If not, `sync.Mutex` should be used because modifying maps are not goroutine-safe.

internal/app/config.go
```diff
@@ -39,6 +40,11 @@ type Auth0Config struct {
 	WebClientID  string
 }
 
+type AuthSrvConfig struct {
+	Domain string `default:"http://localhost:8080/"`
```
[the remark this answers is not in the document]
yk-eukarya: For some reason the login process fails when I remove the slash. Anyway I made this config a strict one and I fixed the issue by manipulate the domain where its used.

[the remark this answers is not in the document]
rot1024: God. it is good to support both domains that contains and does not a slash (if slash exists, trim it)

internal/app/oauth/storage.go
yk-eukarya: The `applicationType` changed into `ApplicationTypeWeb`. 
For the `authMethod` I think its ok to be `AuthMethodNone` as we can consider our front end app as a public client.

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
+package app
```
rot1024, line 1: File name suggestion:

`auth-srv.go` -> `server.go`

- This file is already at "appauth" directory, so it's clear that it is about auth.
- Snakecase is not commonly used as file names in Golang.

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
...
+			if r.Header.Get("Content-Type") != "" {
+				value, _ := header.ParseValueAndParams(r.Header, "Content-Type")
+				if value != "application/json" {
+					// Content-Type header is not application/json
+					handler.ServeHTTP(w, r)
+					return
+				}
+			}
+
+			fmt.Println("JSON -> Form: " + r.URL.String())
```
rot1024, line 96:
```suggestion
```

Do not print content that is not very important. Limiting the content displayed on the console to what is important will help with troubleshooting.

If it is an important log, use `log` package.

internal/app/appauth/sharedClients.go
```diff
@@ -0,0 +1,25 @@
+package appauth
```
rot1024, line 1: File name suggestion:

`sharedClient.go` -> `client.go`

Camelcase is not commonly used as file names in golang. Golang uses only lowercase (also hyphen is not used) as file names.

internal/adapter/http/user.go
```diff
@@ -30,6 +30,11 @@ type CreateUserOutput struct {
 	Email string `json:"email"`
 }
 
+type UserCredentialInput struct {
+	Email    string `json:"email"`
+	Password string `json:"password"`
+}
+
```
rot1024, line 37:
```suggestion
```

Not used

internal/usecase/interactor/user.go
```diff
@@ -164,6 +164,18 @@ func (i *User) Signup(ctx context.Context, inp interfaces.SignupParam) (u *user.
 	return u, team, nil
 }
 
+func (i *User) GetUserByCredentials(ctx context.Context, inp interfaces.GetUserByCredentials) (u *user.User, err error) {
+	u, err = i.userRepo.FindByNameOrEmail(ctx, inp.Email)
+	if err != nil && !errors.Is(err, rerror.ErrNotFound) {
+		return nil, err
+	}
```
rot1024, line 171:
```suggestion
	}
	if u == nil {
		return nil, interfaces.ErrInvalidUserCredentials
	}
```

It's important for security to return the same error even when user is not found.

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
...
+		for _, method := range methods {
+			r.Add(method, path, echo.WrapHandler(route.GetHandler()))
+			fmt.Println(method + "| " + path)
+		}
+
+		return nil
+	}
+}
+
+type Login struct {
```
rot1024, line 142:
```suggestion
type loginForm struct {
```

Private is ok, and name is too abstract.

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
...
+				if value != "application/json" {
+					// Content-Type header is not application/json
+					handler.ServeHTTP(w, r)
+					return
+				}
+			}
+
+			fmt.Println("JSON -> Form: " + r.URL.String())
+			err := r.ParseForm()
+			if err != nil {
```
rot1024, line 98:
```suggestion
			if err := r.ParseForm(); err != nil {
```

Minimize variable scopes

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
...
+				return
+			}
+
+			var result map[string]string
+
+			err = json.NewDecoder(r.Body).Decode(&result)
+			if err != nil {
```
rot1024, line 105:
```suggestion
			if err = json.NewDecoder(r.Body).Decode(&result); err != nil {
```

Minimize variable scopes

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
...
+		op.WithCustomKeysEndpoint(op.NewEndpoint(".well-known/jwks.json")),
+	)
+	if err != nil {
+		e.Logger.Fatal(err)
+	}
+
+	router := handler.HttpHandler().(*mux.Router)
+
+	err = router.Walk(muxToEchoMapper(r))
+	if err != nil {
```
rot1024, line 56:
```suggestion
	if err := router.Walk(muxToEchoMapper(r)); err != nil {
```

Minimize variable scopes

internal/app/auth-srv.go
```diff
@@ -0,0 +1,211 @@
...
+		return
```
rot1024, line 57:
```suggestion
```

Not needed return

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func (s *Storage) SetUserinfoFromScopes(ctx context.Context, userinfo oidc.UserInfoSetter, subject, _ string, _ []string) error {
+
+	request, err := s.AuthRequestBySubject(ctx, subject)
+	if err != nil {
+		return err
+	}
+
+	userinfo.SetSubject(request.GetSubject())
```
rot1024, line 289:
```suggestion
	// TODO: set user info correctly
	userinfo.SetSubject(request.GetSubject())
```

Implement, or mark with a comment

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func (s *Storage) CreateAccessAndRefreshTokens(_ context.Context, request op.TokenRequest, _ string) (accessTokenID string, newRefreshToken string, expiration time.Time, err error) {
+	authReq := request.(*AuthRequest)
+	return "id", authReq.ID, time.Now().UTC().Add(5 * time.Minute), nil
+}
+
+func (s *Storage) TokenRequestByRefreshToken(ctx context.Context, refreshToken string) (op.RefreshTokenRequest, error) {
+	r, err := s.AuthRequestByID(ctx, refreshToken)
+	return r.(op.RefreshTokenRequest), err
```
rot1024, line 222:
```suggestion
	r, err := s.AuthRequestByID(ctx, refreshToken)
	if err != nil {
		return nil, err
	}
	return r.(op.RefreshTokenRequest), nil
```

Note that `r.(xxx)` can occur panic, so this code means r must be always `op.RefreshTokenRequest`

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,347 @@
...
+func initData(s *Storage) {
+
+	client := initLocalClient(s.appConfig.Debug)
+
+	s.clients = map[string]op.Client{
+		client.GetID(): client,
+	}
+
+	s.requests = make(map[string]AuthRequest)
+}
```
rot1024, line 57: Too separated funcs

```suggestion
func NewAuthStorage(cfg *StorageConfig) op.Storage {
	client := initLocalClient(s.appConfig.Debug)
	// generate pkix.Name here from StorageConfig
	name := dummyPkixName
	// if cfg... { ... }

	s := &Storage{
		appConfig: cfg,
		clients: map[string]op.Client{
			client.GetID(): client,
		},
		requests: make(map[string]AuthRequest),
		keySet: initKeys(name),
	}

	return s
}

var dummyPkixName = pkix.Name{
	  Organization:  []string{"Company, INC."},
	  Country:       []string{"US"},
	  Province:      []string{""},
	  Locality:      []string{"San Francisco"},
	  StreetAddress: []string{"Golden Gate Bridge"},
	  PostalCode:    []string{"94016"},
}
```

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,347 @@
...
+
+func initKeys(s *Storage) {
```
rot1024, line 59:
```suggestion
func initKeys(name pkix.Name) (*rsa.PrivateKey, jose.JSONWebKeySet) {
```

The code that causes the side effect may reduce the versatility, maintainability, and testability of the function more than necessary.

In this case, it should be the responsibility of the caller to set the key to the storage. By lines of `s.xxx = xxx`, the versatility, maintainability, and testability of initKeys function is lost.

rot1024 (changes requested):
- Does this PR not include mongo implementation?
- DO `go mod tidy` to resolve error from golangci-lint

internal/app/auth-srv.go
yk-eukarya: The file is not in the `appauth` folder that because of import cycle. anyway I've re name it to `auth_server`.

yk-eukarya:
This PR does not contains mongo implementation, I'm planning to make with another PR.
`go mod tidy` did not resolve the error from `golangci-lint` !

rot1024:
Commit go.mod

yk-eukarya:
I did commit the `go.mod` and `go.sum` files

internal/app/appauth/storage.go
```diff
@@ -33,48 +33,44 @@ type StorageConfig struct {
...
-	initData(s)
-	initKeys(s)
+func NewAuthStorage(cfg *StorageConfig) op.Storage {
 
-	return s
-}
+	client := initLocalClient(cfg.Debug)
 
-func initData(s *Storage) {
+	name := dummyPkixName
```
rot1024, line 49: Generating pkix.Name from config is not implemented yet?

internal/usecase/interactor/user.go
```diff
@@ -166,8 +166,8 @@ func (i *User) Signup(ctx context.Context, inp interfaces.SignupParam) (u *user.
 
 func (i *User) GetUserByCredentials(ctx context.Context, inp interfaces.GetUserByCredentials) (u *user.User, err error) {
 	u, err = i.userRepo.FindByNameOrEmail(ctx, inp.Email)
-	if err != nil && !errors.Is(err, rerror.ErrNotFound) {
-		return nil, err
+	if err != nil || u == nil {
+		return nil, interfaces.ErrInvalidUserCredentials
```
rot1024, line 170: I did not say errors from the DB should be erased. Internal errors returned by the handler are automatically output to the standard output, so it is necessary to return the internal errors of the DB in order to determine the cause.

```suggestion
	if err != nil && !errors.Is(rerror.ErrNotFound, err) {
		return nil, err
	} else if u == nil {
		return nil, interfaces.ErrInvalidUserCredentials
	}
```

rot1024 (changes requested):

internal/app/config.go
```diff
@@ -39,6 +40,21 @@ type Auth0Config struct {
 	WebClientID  string
 }
 
+type AuthConfig struct {
+	Domain string `default:"http://localhost:8080"`
+	Key    string
+	Pkix   AuthPkixConfig
+}
+
+type AuthPkixConfig struct {
```
rot1024, line 49: DN is better

- DN = Distinguished name (X.509 DN)
- PKIX = Public Key Infrastructure working group

```suggestion
type AuthDNConfig struct {
```

internal/app/config.go
```diff
@@ -39,6 +40,21 @@ type Auth0Config struct {
 	WebClientID  string
 }
 
+type AuthConfig struct {
+	Domain string `default:"http://localhost:8080"`
+	Key    string
+	Pkix   AuthPkixConfig
```
rot1024, line 46:
```suggestion
	DN   *AuthDNConfig
```

Make the config optional to use dummy data if there is no DN config

internal/app/config.go
```diff
@@ -39,6 +40,21 @@ type Auth0Config struct {
...
+}
+
+type AuthPkixConfig struct {
+	Organization  string `default:"Company, INC."`
+	Country       string `default:"US"`
+	Province      string `default:""`
+	Locality      string `default:"San Francisco"`
+	StreetAddress string `default:"Golden Gate Bridge"`
+	PostalCode    string `default:"94016"`
```
rot1024, line 55: DN can accept multi strings

```suggestion
	Organization  []string
	Country       []string
	Province      []string
	Locality      []string
	StreetAddress []string
	PostalCode    []string
```

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+type AuthPkixConfig struct {
+	Organization  string
+	Country       string
+	Province      string
+	Locality      string
+	StreetAddress string
+	PostalCode    string
+}
```
rot1024, line 44:
```suggestion
	DN   *pkix.Name
}
```

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,335 @@
...
+	client := initLocalClient(cfg.Debug)
+
+	name := pkix.Name{
+		Organization:  []string{cfg.Pkix.Organization},
+		Country:       []string{cfg.Pkix.Country},
+		Province:      []string{cfg.Pkix.Province},
+		Locality:      []string{cfg.Pkix.Locality},
+		StreetAddress: []string{cfg.Pkix.StreetAddress},
+		PostalCode:    []string{cfg.Pkix.PostalCode},
+	}
```
rot1024, line 57: Use dummy data here

```suggestion
	name := dummyName
	if cfg.DN != nil {
		name = *cfg.DN
	}
```

rot1024 (changes requested):
Using dummy values for config defaults is not a good idea because if you specify only a part of the DN, the rest of the fields will be set to dummy values.

rot1024:
How to manage the root CA:

- Using managed service: e.g. GCP Certificate Authority Service
- Manually deploy certs
   - Build PKI with HashiCorp Vault and issue certs with it
- Manually generate (Now we are supporting)
   - Certificate revocation lists are cumbersome to manage
   - Care must be taken to manage the private key of the root CA

Therefore, in the future, it may be necessary to further consider how to deploy multiple certificates.

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,335 @@
...
+
+func initKeys(name pkix.Name) (*rsa.PrivateKey, jose.JSONWebKeySet) {
+
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err)
+	}
+
+	cert := &x509.Certificate{
+		SerialNumber: big.NewInt(1658),
```
rot1024, line 80: Set serial number that is clearly dummy number

```suggestion
		SerialNumber: big.NewInt(1),
```

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,335 @@
...
+		Subject:      name,
+		IPAddresses:  []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback},
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(10, 0, 0),
+		IsCA:         true,
+		SubjectKeyId: []byte{1, 2, 3, 4, 6},
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
```
rot1024, line 87: This cert is not for server cert, so ExtKeyUsage is unnecessary

```suggestion
```

internal/app/appauth/storage.go
```diff
@@ -0,0 +1,335 @@
...
+
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err)
+	}
+
+	cert := &x509.Certificate{
+		SerialNumber: big.NewInt(1658),
+		Subject:      name,
+		IPAddresses:  []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback},
```
rot1024, line 82:
```suggestion
```

Not nesessary

internal/app/appauth/client.go
```diff
@@ -0,0 +1,98 @@
...
+	applicationType    op.ApplicationType
+	authMethod         oidc.AuthMethod
+	accessTokenType    op.AccessTokenType
+	responseTypes      []oidc.ResponseType
+	grantTypes         []oidc.GrantType
+	allowedScopes      []string
+	redirectURIs       []string
+	logoutRedirectURIs []string
+	loginURI           string
+	iDTokenLifetime    time.Duration
```
rot1024, line 22:
```suggestion
	idTokenLifetime    time.Duration
```

rot1024 (changes requested):
I have researched X.509 and reviewed the PR further.

rot1024 (changes requested):

internal/app/appauth/clients.go
```diff
@@ -0,0 +1,25 @@
...
+		responseTypes:   []oidc.ResponseType{oidc.ResponseTypeCode},
+		grantTypes:      []oidc.GrantType{oidc.GrantTypeCode, oidc.GrantTypeRefreshToken},
+		redirectURIs:    []string{"http://localhost:3000"},
+		allowedScopes:   []string{"openid", "profile", "email"},
+		loginURI:        "http://localhost:3000/login?id=%s",
+		idTokenLifetime: 5 * time.Minute,
+		clockSkew:       0,
+		devMode:         devMode,
+	}
+}
```
rot1024, line 25: This code can be move to client.go and clients.go can be  deleted.

internal/app/config.go
```diff
@@ -39,6 +40,23 @@ type Auth0Config struct {
 	WebClientID  string
 }
 
+type AuthConfig struct {
+	Domain string `default:"http://localhost:8080"`
+	Key    string
```
rot1024, line 45: Not used
```suggestion
```

internal/app/auth_server.go
```diff
@@ -0,0 +1,226 @@
...
+		methods, err := route.GetMethods()
+		if err != nil {
+			r.Any(path, echo.WrapHandler(route.GetHandler()))
+			fmt.Println("ANY| " + path)
+			return nil
+		}
+
+		for _, method := range methods {
+			r.Add(method, path, echo.WrapHandler(route.GetHandler()))
+			fmt.Println(method + "| " + path)
```
rot1024, line 150: Replace ALL fmt.Println to log.Infof and add prefix to log message: "auth"
```suggestion
			log.Infof("auth: %s %s", method, path)
```

Note: log is from reearth-backend/pkg/log

internal/app/auth_server.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	}
+
+	storage := appauth.NewAuthStorage(
+		&appauth.StorageConfig{
+			Domain: domain.String(),
+			Debug:  cfg.Debug,
+			DN:     dn,
+		},
+		userUsecase.GetUserBySubject,
```
rot1024, line 58:
```suggestion
		func (ctx context.Contrxt, subject string) (*user.User, error) {
  return userUsecase.GetUserBySubject(ctx, subject, getOperator(ctx))
},
```

Operator is required

rot1024 (changes requested):

internal/app/config.go
yk-eukarya: It's used [here ](https://github.com/reearth/reearth-backend/blob/f493cd30062bf473703ba78ef50cb5a39cd5c979/internal/app/auth_server.go#L34)

internal/app/auth_server.go
yk-eukarya: Those ware for debugging, I removed them

internal/app/auth_server.go
yk-eukarya: since our end points here are not authenticated, how to use the operator?
the only exception is `userinfo` endpoint and here the `caos/oidc` lib validates the access token.

internal/app/config.go
rot1024: Sorry i forgot it

internal/app/auth_server.go
rot1024: Ok right, operator can be nil. But the usecase func accepts operator as third arg, so it is good to wraps it with new func to clearly show operator is nil, or delete the third arg from the usecase func if operator is not used in the func.

yk-eukarya: I've already removed the operator from the usecase func, that's because its not used

rot1024 (approved):

PR #79, opened 2021-11-19
feat: password reset

yk-eukarya:


internal/adapter/http/user.go
```diff
@@ -17,6 +17,15 @@ func NewUserController(usecase interfaces.User) *UserController {
 	}
 }
 
+type PasswordResetRequestInput struct {
+	Email *string `json:"email"`
```
rot1024, line 21:
```suggestion
	Email string `json:"email"`
```

Email is required, so it does not need to be pointer

internal/adapter/http/user.go
```diff
@@ -17,6 +17,15 @@ func NewUserController(usecase interfaces.User) *UserController {
...
+}
+
+type PasswordResetConfirmInput struct {
+	Token    *string `json:"token"`
+	Password *string `json:"password"`
```
rot1024, line 26:
```suggestion
	Token    string `json:"token"`
	Password string `json:"password"`
```

internal/app/public.go
```diff
@@ -39,6 +39,34 @@ func publicAPI(
 		return c.JSON(http.StatusOK, output)
 	})
 
+	r.POST("/password-reset-request", func(c echo.Context) error {
+		var inp http1.PasswordResetRequestInput
+		if err := c.Bind(&inp); err != nil {
+			return &echo.HTTPError{Code: http.StatusBadRequest, Message: fmt.Errorf("failed to parse request body: %w", err)}
```
rot1024, line 45: As you can see https://github.com/labstack/echo/blob/master/bind.go#L59, errors do not need to be wrapped with echo.HTTPError (they are already wrapped)

```suggestion
			return err
```

internal/app/public.go
```diff
@@ -39,6 +39,34 @@ func publicAPI(
 		return c.JSON(http.StatusOK, output)
 	})
 
+	r.POST("/password-reset-request", func(c echo.Context) error {
```
rot1024, line 42:
```suggestion
	r.POST("/password", func(c echo.Context) error {
```

Simple URL is good

internal/app/public.go
```diff
@@ -39,6 +39,34 @@ func publicAPI(
...
+
+		output, err := controller.PasswordResetRequest(c.Request().Context(), inp)
+		if err != nil {
+			return err
+		}
+
+		return c.JSON(http.StatusOK, output)
+	})
+
+	r.POST("/password-reset-confirm", func(c echo.Context) error {
```
rot1024, line 56:
```suggestion
	r.POST("/password-reset", func(c echo.Context) error {
```

Also you can consider combining 2 small APIs into `POST /password` as following:

1. If body contains email, send password reset mail
2. If body contains token and password, reset password
3. Else, return error

internal/infrastructure/mongo/mongodoc/user.go
```diff
@@ -58,6 +69,12 @@ func NewUser(user *user1.User) (*UserDocument, string) {
 		Lang:         user.Lang().String(),
 		Theme:        string(user.Theme()),
 		Password:     user.Password(),
+		PasswordReset: &PasswordResetDocument{
+			Token:     pwdReset.Token,
+			CreatedAt: pwdReset.CreatedAt,
+			ExpiresAt: pwdReset.ExpiresAt,
+			IsUsed:    pwdReset.IsUsed,
+		},
```
rot1024, line 77: If pwdReset is nil, PasswordReset field in UserDocument should be nil, otherwise the app will be panic

internal/infrastructure/mongo/mongodoc/user.go
```diff
@@ -86,6 +103,7 @@ func (d *UserDocument) Model() (*user1.User, error) {
 		LangFrom(d.Lang).
 		Password("", d.Password).
 		Theme(user.Theme(d.Theme)).
+		PasswordReset(d.PasswordReset.Token, d.PasswordReset.IsUsed, d.PasswordReset.ExpiresAt, d.PasswordReset.CreatedAt).
```
rot1024, line 106: Ifd.PasswordReset is nil, PasswordReset field in User should be nil, otherwise the app will be panic

internal/usecase/interactor/user.go
```diff
@@ -237,6 +237,61 @@ func (i *User) GetUserBySubject(ctx context.Context, sub string) (u *user.User,
 	return u, nil
 }
 
+func (i *User) PasswordResetRequest(ctx context.Context, param interfaces.PasswordResetRequestParam) (bool, error) {
```
rot1024, line 240:
```suggestion
func (i *User) StartPasswordReset(ctx context.Context, email string) error {
```

Only error is enough. If no error returns, this operation was succeeded. Also param struct does not need because there are few params.

internal/usecase/interactor/user.go
```diff
@@ -237,6 +237,61 @@ func (i *User) GetUserBySubject(ctx context.Context, sub string) (u *user.User,
...
+
+	tx, err := i.transaction.Begin()
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		if err2 := tx.End(ctx); err == nil && err2 != nil {
+			err = err2
+		}
+	}()
```
rot1024, line 270: Transaction should be started at beginning of usecase

internal/usecase/interactor/user.go
```diff
@@ -237,6 +237,61 @@ func (i *User) GetUserBySubject(ctx context.Context, sub string) (u *user.User,
...
+
+	err = i.userRepo.Save(ctx, u)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (i *User) PasswordResetConfirm(ctx context.Context, param interfaces.PasswordResetConfirmParam) (bool, error) {
```
rot1024, line 256:
```suggestion
func (i *User) PasswordReset(ctx context.Context, token, password string) error {
```

pkg/user/password_rest.go
```diff
@@ -0,0 +1,38 @@
...
+		Token:     generateToken(),
+		CreatedAt: time.Now(),
+		ExpiresAt: time.Now().Add(24 * time.Hour),
+		IsUsed:    false,
+	}
+}
+
+func generateToken() string {
+	t := time.Now()
+	entropy := ulid.Monotonic(rand.New(rand.NewSource(t.UnixNano())), 0)
```
rot1024, line 28: I think UUIDv4 is better because advantages of ULID don't mean much this usecase.

ULID adbantages:
- Short
- URL safe -> No sense for this usecase
- Timestamp is embed and sortable -> No sense for this usecase

ULID weakness:
- Short so easy to guess -> I'm somewhat nervous about using ULID for this use case.
- Timestamp ca be read by anyone
- Clocks must be synchronized to be used in distributed systems

pkg/user/password_rest.go
```diff
@@ -0,0 +1,38 @@
...
+	return ulid.MustNew(ulid.Timestamp(t), entropy).String()
+}
+
+func (pr *PasswordReset) IsValidRequest(token string) bool {
+	return !pr.IsUsed && pr.Token == token && time.Now().Before(pr.ExpiresAt)
```
rot1024, line 33: Add uni test. Other methods are also.

```suggestion
	return pr != nil && !pr.IsUsed && pr.Token == token && time.Now().Before(pr.CreatedAt.Add(24 * time.Hour))
```

pkg/user/password_rest.go
```diff
@@ -0,0 +1,38 @@
...
+	"math/rand"
+	"time"
+
+	"github.com/oklog/ulid"
+)
+
+type PasswordReset struct {
+	Token     string
+	CreatedAt time.Time
+	ExpiresAt time.Time
```
rot1024, line 13: No need because just adding 24 hours to CreatedAt is enough to detect it is expired
```suggestion
```

pkg/user/password_rest.go
```diff
@@ -0,0 +1,38 @@
...
+	IsUsed    bool
```
rot1024, line 14: Instead of IsUsed, how about the approach of deleting the entire PasswordReset after the password reset is complete? This will make PasswordReset struct immutable

rot1024 (changes requested):

internal/adapter/http/user.go
```diff
@@ -17,6 +17,12 @@ func NewUserController(usecase interfaces.User) *UserController {
 	}
 }
 
+type PasswordResetInput struct {
+	Email    *string `json:"email"`
+	Token    *string `json:"token"`
+	Password *string `json:"password"`
```
rot1024, line 23:
```suggestion
	Email    string `json:"email"`
	Token    string `json:"token"`
	Password string `json:"password"`
```

Empty string is almost same as nil

internal/app/public.go
```diff
@@ -39,6 +39,31 @@ func publicAPI(
...
+			err := controller.StartPasswordReset(c.Request().Context(), inp)
+			if err != nil {
+				return err
+			}
+			return c.JSON(http.StatusOK, true)
+		}
+
+		if inp.Token != nil && inp.Password != nil {
+			err := controller.PasswordReset(c.Request().Context(), inp)
+			if err != nil {
```
rot1024, line 58:
```suggestion
			if err := controller.PasswordReset(c.Request().Context(), inp); err != nil {
```

internal/app/public.go
```diff
@@ -39,6 +39,31 @@ func publicAPI(
...
 
+	r.POST("/password-reset", func(c echo.Context) error {
+		var inp http1.PasswordResetInput
+		if err := c.Bind(&inp); err != nil {
+			return err
+		}
+
+		if inp.Email != nil {
+			err := controller.StartPasswordReset(c.Request().Context(), inp)
+			if err != nil {
```
rot1024, line 50:
```suggestion
			if err := controller.StartPasswordReset(c.Request().Context(), inp); err != nil {
```

internal/usecase/interactor/user.go
```diff
@@ -237,6 +237,61 @@ func (i *User) GetUserBySubject(ctx context.Context, sub string) (u *user.User,
 	return u, nil
 }
 
+func (i *User) StartPasswordReset(ctx context.Context, email string) error {
+	u, err := i.userRepo.FindByEmail(ctx, email)
```
rot1024, line 241: Transaction is required

internal/usecase/interactor/user.go
```diff
@@ -237,6 +237,61 @@ func (i *User) GetUserBySubject(ctx context.Context, sub string) (u *user.User,
...
+	if err != nil {
+		return err
+	}
+
+	u.CreatePasswordReset()
+
+	err = i.userRepo.Save(ctx, u)
+	if err != nil {
```
rot1024, line 249:
```suggestion
	if err := i.userRepo.Save(ctx, u); err != nil {
```

pkg/user/builder.go
```diff
@@ -92,3 +94,11 @@ func (b *Builder) Auths(auths []Auth) *Builder {
 	b.u.auths = append([]Auth{}, auths...)
 	return b
 }
+
+func (b *Builder) PasswordReset(token string, createdAt time.Time) *Builder {
```
rot1024, line 98:
```suggestion
func (b *Builder) PasswordReset(pr *PasswordReset) *Builder {
```

Accepting PasswordReset struct directly make codes simple

pkg/user/password_reset.go
```diff
@@ -0,0 +1,27 @@
...
+	Token     string
+	CreatedAt time.Time
+}
+
+func NewPasswordReset() *PasswordReset {
+	return &PasswordReset{
+		Token:     generateToken(),
+		CreatedAt: time.Now(),
+	}
+}
```
rot1024, line 19: You can add another func for conversion between domain model and mongo doc:

```go
func PasswordResetFrom(token string, createdAt time.Time) *PasswordReset {
	return &PasswordReset{
		Token:     token,
		CreatedAt: createdAt,
	}
}
```

Of course, unit test is required

pkg/user/password_reset.go
```diff
@@ -0,0 +1,27 @@
...
+
+func generateToken() string {
+	return uuid.UUIDv4()
+}
+
+func (pr *PasswordReset) IsValidRequest(token string) bool {
```
rot1024, line 25:
```suggestion
func (pr *PasswordReset) Valid(token string) bool {
```

Short method name will be liked by Go developers

pkg/user/user.go
```diff
@@ -178,3 +179,15 @@ func verifyPassword(toVerify string, encoded []byte) (bool, error) {
...
+	return u.passwordReset
+}
+
+func (u *User) CreatePasswordReset() {
+	u.passwordReset = NewPasswordReset()
+}
+
+func (u *User) RemovePasswordReset() {
+	u.passwordReset = nil
+}
```
rot1024, line 193:
```suggestion
func (u *User) SetPasswordReset(pr *PasswordReset) {
	u.passwordReset = pr.Clone()
}
```

This implementation is one of answer.

My way is using setter that simplifies User struct because the internal concerns of password reset can be completely transferred to PasswordReset struct. Copying is required to prevent changing internal values by outside without methods.

rot1024 (changes requested):

pkg/user/builder.go
yk-eukarya: @rot1024 
I tried this one it cause an import loop!

pkg/user/password_reset.go
```diff
@@ -0,0 +1,42 @@
+package user
+
+import (
+	"time"
+
+	"github.com/dgryski/trifles/uuid"
```
rot1024, line 6: Use `github.com/google/uuid` (use a well-maintained library for security reasons)

pkg/user/password_reset_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewPasswordReset(t *testing.T) {
+	pr := NewPasswordReset()
+	assert.NotNil(t, pr)
+	assert.NotEmpty(t, pr.Token)
+	assert.True(t, time.Until(pr.CreatedAt) < 1*time.Second)
```
rot1024, line 14: You can use https://pkg.go.dev/github.com/stretchr/testify/assert#Less

pkg/user/password_reset_test.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func TestPasswordReset_IsValidRequest(t *testing.T) {
```
rot1024, line 17:
```suggestion
func TestPasswordReset_Validate(t *testing.T) {
```

pkg/user/password_reset.go
```diff
@@ -0,0 +1,42 @@
...
+		Token:     token,
+		CreatedAt: createdAt,
+	}
+}
+
+func generateToken() string {
+	return uuid.UUIDv4()
+}
+
+func (pr *PasswordReset) Valid(token string) bool {
```
rot1024, line 32: Sorry, I made a mistake and thought this method was just a simple getter, but it accepts an arg and has logic. I think `Validate` is better name

```suggestion
func (pr *PasswordReset) Validate(token string) bool {
```

rot1024 (changes requested):

pkg/user/password_reset_test.go
yk-eukarya: Less can not compare type "time.Time"

rot1024:
```suggestion
	pr := NewPasswordReset()
	assert.NotNil(t, pr)
	assert.NotEmpty(t, pr.Token)
	assert.Less(t, int64(time.Until(pr.CreatedAt)), int64(1*time.Second))
```

rot1024 (changes requested):
Do not edit `go.mod` manually. Use `go get xxx` to add new package. Please exec `go mod tidy` to make go.mod clean

go.mod
```diff
@@ -69,6 +69,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 )
 
+require github.com/google/uuid v1.3.0
+
 require (
```
rot1024, line 74:
```suggestion
	github.com/google/uuid v1.3.0
```

rot1024 (changes requested):
Do not edit go.mod manually. Use `go get xxx` command to add new package. 

- Fix go.mod following my suggestion
- Exec `go mod tidy` to sort the requires nicely

pkg/user/password_reset_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewPasswordReset(t *testing.T) {
+	pr := NewPasswordReset()
+	assert.NotNil(t, pr)
+	assert.NotEmpty(t, pr.Token)
+	time.Sleep(1 * time.Second)
+	assert.Less(t, pr.CreatedAt.Unix(), time.Now().Unix())
```
rot1024, line 15: Don't use sleep in the test. It make unit tests slow. Consider mocking `time.Now`.

```go
var timeNow = time.Now

func NewPasswordReset() *PasswordReset {
	return &PasswordReset{
		// ...
		CreatedAt: timeNow(),
	}
}
```

```suggestion
	mockTime := time.Now()
	timeNow = func() time.Time { return mockTime }
	pr := NewPasswordReset()
	assert.NotNil(t, pr)
	assert.NotEmpty(t, pr.Token)
	assert.Equal(t, mockTime, pr.CreatedAt)
```

rot1024 (changes requested):

pkg/user/password_reset_test.go
yk-eukarya: @rot1024 In this particular case I could solve it without sleep nor mock

rot1024 (approved):

PR #81, opened 2021-11-22
feat: asset filtering

KaWaite:
# Overview
Adding infinite scroll to assets means we need to move filter/sorting and searching of assets to the backend during pagination. 

## What I've done
- Updated graphql query to include sort type
- Add sorting logic to asset querying
- Paginated search results for asset name

## Memo
- I have made it so if the query receives `before` and `last` it'll be reversed, or `first` and `after` the results will be in normal order. Front-end will pass the appropriate params depending on if the user selects reverse order or not


internal/infrastructure/mongo/mongodoc/client.go
```diff
@@ -165,42 +165,34 @@ func getCursor(raw bson.Raw, key string) (*usecase.Cursor, error) {
...
+	count, err := coll.CountDocuments(ctx, filter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to count documents: %v", err.Error())
 	}
-	if !paginate {
-		return nil, fmt.Errorf("missing pagination paramiters: %v", p)
+
+	limit, op, cur, err := p.Parameters()
+	if err != nil {
+		return nil, fmt.Errorf("faild to parse pagination parameters: %w", err)
```
KaWaite, line 195: failed, not faild

internal/infrastructure/mongo/mongodoc/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+		After:  (*string)(pagination.After),
+		First:  pagination.First,
+		Last:   pagination.Last,
+	}
+}
+
+func (p *Pagination) SortDirection() int {
+	if p != nil {
+		return 1
+	}
```
KaWaite, line 31: Should this not be:
```
if p.First != nil {
    return 1
}
```

internal/infrastructure/mongo/mongodoc/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+		cursor = after
+		return
+	}
+	if last, before := p.Last, p.Before; last != nil {
+		limit = int64(*last)
+		op = "$lt"
+		cursor = before
+		return
+	}
+	return 0, "", nil, errors.New("neither first nor last are not set")
```
KaWaite, line 51: neither first nor last are set
OR
either first or last are not set

internal/adapter/gql/loader_asset.go
```diff
@@ -32,9 +33,10 @@ func (c *AssetLoader) Fetch(ctx context.Context, ids []id.AssetID) ([]*gqlmodel.
 	return assets, nil
 }
 
-func (c *AssetLoader) FindByTeam(ctx context.Context, teamID id.ID, first *int, last *int, before *usecase.Cursor, after *usecase.Cursor) (*gqlmodel.AssetConnection, error) {
-	p := usecase.NewPagination(first, last, before, after)
-	assets, pi, err := c.usecase.FindByTeam(ctx, id.TeamID(teamID), p, getOperator(ctx))
+func (c *AssetLoader) FindByTeam(ctx context.Context, teamID id.ID, keyword *string, sort *asset.SortType, pagination *gqlmodel.Pagination) (*gqlmodel.AssetConnection, error) {
+	p := gqlmodel.ToPagination(pagination)
+
+	assets, pi, err := c.usecase.FindByTeam(ctx, id.TeamID(teamID), keyword, (*string)(sort), p, getOperator(ctx))
```
rot1024, line 39:
```suggestion
	assets, pi, err := c.usecase.FindByTeam(ctx, id.TeamID(teamID), keyword, gqlmodel.AssetSortTypeFrom(sort), p, getOperator(ctx))
```

internal/usecase/repo/asset.go
```diff
@@ -11,7 +11,7 @@ import (
 type Asset interface {
 	Save(context.Context, *asset.Asset) error
 	Remove(context.Context, id.AssetID) error
-	FindByTeam(context.Context, id.TeamID, *usecase.Pagination) ([]*asset.Asset, *usecase.PageInfo, error)
+	FindByTeam(context.Context, id.TeamID, *string, *string, *usecase.Pagination) ([]*asset.Asset, *usecase.PageInfo, error)
```
rot1024, line 14: It is very confusing so using a struct is good:

```go
// defined at internal/usecase/repo/asset.go
type AssetFilter struct {
	Sort       *asset.SortType
	Keyword    *string
	Pagination *usecase.Pagination
}
```

```suggestion
	FindByTeam(context.Context, id.TeamID, AssetFilter) ([]*asset.Asset, *usecase.PageInfo, error)
```

rot1024 (changes requested):

internal/infrastructure/mongo/mongodoc/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/reearth/reearth-backend/internal/usecase"
+)
+
+type Pagination struct {
+	Before *string
+	After  *string
+	First  *int
+	Last   *int
+}
```
rot1024, line 14: ✅  Excellent!
Understanding that pagination for MongoDB is different from pagination for the use case layer, and separating them into separate structs to achieve separation of interest is a good point worth mentioning.

internal/infrastructure/mongo/mongodoc/client.go
```diff
@@ -202,45 +202,70 @@ func getCursor(raw bson.Raw, key string) (*usecase.Cursor, error) {
...
+						key:     bson.M{"$gt": *cur},
+					},
+				},
+			}
 		}
-		if before := p.Before; before != nil {
-			filter = appendE(filter, bson.E{Key: key, Value: bson.D{
-				{Key: "$lt", Value: *before},
-			}})
+
```
rot1024, line 259: Processing to build paginationFilter is very long, so separate it to another function. It is hard to read codes

rot1024 (changes requested):

internal/infrastructure/mongo/mongodoc/client.go
```diff
@@ -312,6 +290,57 @@ func (c *Client) Paginate(ctx context.Context, col string, filter interface{}, p
...
+	var sortKey = ""
+	if sort != nil && len(*sort) > 0 && *sort != "id" {
+		sortKey = *sort
+		sortOptions = append(sortOptions, bson.E{Key: sortKey, Value: p.SortDirection()})
+	}
+	sortOptions = append(sortOptions, bson.E{Key: key, Value: p.SortDirection()})
+	return sortOptions, sortKey
+}
+
+func paginationFilter(ctx context.Context, coll *mongo.Collection, p *Pagination, sortKey, key string, filter bson.M) (bson.M, *int64, error) {
```
rot1024, line 304: When a function returns a reference, it means that the responsibility for its nil check is delegated to the caller of the function. Forgetting this can lead to panic. In this case, it is not necessary that the limit be a reference, since the limit will simply be ignored if an err is encountered.

```suggestion
func paginationFilter(ctx context.Context, coll *mongo.Collection, p *Pagination, sortKey, key string, filter bson.M) (bson.M, int64, error) {
```

internal/infrastructure/mongo/mongodoc/client.go
```diff
@@ -312,6 +290,57 @@ func (c *Client) Paginate(ctx context.Context, col string, filter interface{}, p
...
+	limit, op, cur, err := p.Parameters()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to parse pagination parameters: %w", err)
```
rot1024, line 307:
```suggestion
		return nil, 0, fmt.Errorf("failed to parse pagination parameters: %w", err)
```

internal/infrastructure/mongo/mongodoc/client.go
```diff
@@ -312,6 +290,57 @@ func (c *Client) Paginate(ctx context.Context, col string, filter interface{}, p
...
+		}
+
+		filter = bson.M{
+			"$and": []bson.M{
+				filter,
+				paginationFilter,
+			},
+		}
+	}
+	return filter, &limit, nil
```
rot1024, line 341:
```suggestion
	return filter, limit, nil
```

rot1024 (changes requested):

internal/infrastructure/mongo/mongodoc/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewPaginationFrom(pagination *usecase.Pagination) *Pagination {
```
rot1024, line 16: should be NewXXX or XXXFrom

internal/infrastructure/mongo/mongodoc/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+		After:  (*string)(pagination.After),
+		First:  pagination.First,
+		Last:   pagination.Last,
+	}
+}
+
+func (p *Pagination) SortDirection() int {
+	if p == nil {
+		return 1
+	}
```
rot1024, line 31: This if block is unnecessary

rot1024 (approved):

internal/infrastructure/mongo/mongodoc/pagination.go
eng-ymk: Resolved

rot1024 (changes requested):
@KaWaite found bugs:

- SIZE and NAME sorting in reverse is messing up.
- First query seems okay. If I ask for 20, it gives me the sorted 20.
- But the second query, so for example ..., sort: SIZE, pagination: {last: 20, before: "EndCursor"} returns for me 19 of the original 20 sorted results.

rot1024 (approved):

PR #82, opened 2021-11-23
feat: Auth logout

yk-eukarya:


internal/app/auth_server.go
```diff
@@ -18,6 +18,12 @@ import (
 	"github.com/golang/gddo/httputil/header"
 )
 
+var (
+	loginEndpoint  = "api/login"
+	logoutEndpoint = "v2/logout"
```
rot1024, line 23:
```suggestion
	logoutEndpoint = "api/logout"
```

internal/app/auth_server.go
```diff
@@ -199,6 +208,13 @@ func login(ctx context.Context, cfg *ServerConfig, storage op.Storage, userUseca
 	}
 }
 
+func logout(_ context.Context, _ *ServerConfig, _ op.Storage, _ interfaces.User) func(ctx echo.Context) error {
+	return func(ec echo.Context) error {
+		u := ec.QueryParam("returnTo")
+		return ec.Redirect(http.StatusFound, u)
```
rot1024, line 214:
```suggestion
		return ec.Redirect(http.StatusTemporaryRedirect, u)
```

internal/app/auth_server.go
```diff
@@ -199,6 +208,13 @@ func login(ctx context.Context, cfg *ServerConfig, storage op.Storage, userUseca
 	}
 }
 
+func logout(_ context.Context, _ *ServerConfig, _ op.Storage, _ interfaces.User) func(ctx echo.Context) error {
+	return func(ec echo.Context) error {
```
rot1024, line 212: This method is nothing to do?

rot1024 (changes requested):

internal/app/auth_server.go
yk-eukarya: @rot1024 
I made the logout endpoint like this because I could not find a way to change it in `auth0/auth0-react` lib, so should I leave it as it is, or change it to `api/logout`

internal/app/auth_server.go
yk-eukarya: I made like this just in case we decided to implement a sessions for the auth server.

internal/app/auth_server.go
rot1024: I see. It is hard-coded: https://github.com/auth0/auth0-spa-js/blob/master/src/Auth0Client.ts#L976

So I suggest create 2 endpoints that use the same handler: one is `api/logout` (for general) and another is `v2/logout` (for Auth0)

internal/app/auth_server.go
rot1024: Great

rot1024 (approved):

PR #83, opened 2021-11-28
feat: user verification

mimoham24:
# Overview

## What I haven't done
create email content


pkg/user/verification_test.go
```diff
@@ -0,0 +1,210 @@
...
+import (
+	"regexp"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewVerification(t *testing.T) {
+
```
rot1024, line 12:
```suggestion
```

pkg/user/user.go
```diff
@@ -59,6 +60,10 @@ func (u *User) UpdateTheme(t Theme) {
 	u.theme = t
 }
 
+func (u *User) Verification() *Verification {
+	return u.verification
```
rot1024, line 64: add unit test for Verification method

pkg/user/user_test.go
```diff
@@ -315,3 +316,15 @@ func TestUser_GetAuthByProvider(t *testing.T) {
...
+func TestUser_SetVerification(t *testing.T) {
+	var input *User
+	input = &User{}
+	v := &Verification{
+		verified:   false,
+		code:       "xxx",
+		expiration: time.Time{},
+	}
+	input.SetVerification(v)
+	assert.Equal(t, v, input.Verification())
```
rot1024, line 329:
```suggestion
	assert.Equal(t, v, input.verification)
```

This test is not for `Verification` method, so it should be passed even if there is no `Verification` method.

Also, if there is some special logic in `Verification` method, the value assigned by `SetVerification` method may change through `Verification` method, which makes it confusing to know what is the behavior of `SetVerification` in this test. Since all `SetVerification` does is assign to a private field, it is simplest to check the side effect of the method, i.e., the value of that private field. is the simplest.

pkg/user/builder.go
```diff
@@ -76,3 +76,8 @@ func (b *Builder) Auths(auths []Auth) *Builder {
 	b.u.auths = append([]Auth{}, auths...)
 	return b
 }
+
+func (b *Builder) Verification(v *Verification) *Builder {
+	b.u.verification = v
```
rot1024, line 81: unit test missing

internal/usecase/interfaces/user.go
```diff
@@ -44,6 +44,8 @@ type UpdateMeParam struct {
 type User interface {
 	Fetch(context.Context, []id.UserID, *usecase.Operator) ([]*user.User, error)
 	Signup(context.Context, SignupParam) (*user.User, *user.Team, error)
+	CreateVerification(context.Context, string) (string, error)
```
rot1024, line 47:
```suggestion
	CreateVerification(context.Context, string) error
```

string return arg is not used. because it is enough to detect whether the sending email is succeeded by checking error is nil

internal/infrastructure/mongo/mongodoc/user.go
```diff
@@ -56,6 +65,11 @@ func NewUser(user *user1.User) (*UserDocument, string) {
 		Team:         user.Team().String(),
 		Lang:         user.Lang().String(),
 		Theme:        string(user.Theme()),
+		Verification: &UserVerificationDoc{
+			Code:       user.Verification().Code(),
+			Expiration: user.Verification().Expiration(),
+			Verified:   user.Verification().IsVerified(),
+		},
```
rot1024, line 72: Verification may not always be present; if it is nil, do not simply create a `UserVerificationDoc`.

internal/app/public.go
```diff
@@ -39,6 +39,33 @@ func publicAPI(
...
+		}
+		output, err := controller.CreateVerification(c.Request().Context(), inp)
+		if err != nil {
+			return err
+		}
+
+		return c.JSON(http.StatusOK, output)
+	})
+
+	r.GET("/signup/verify/:code", func(c echo.Context) error {
```
rot1024, line 55:
```suggestion
	r.POST("/signup/verify/:code", func(c echo.Context) error {
```

- HTTP `GET` should not cause any side effects
- Often, when you click the link on the email, you will see a screen saying that the verification is complete, but let's leave this to the frontend. The backend simply returns JSON, and the frontend calls this API when the screen is displayed. (On the other hand, if you want to display this screen in the backend, you need to use GET so that the browser can display it. This is a special case of GET that causes side effects)

rot1024 (changes requested):

rot1024 (changes requested):

pkg/user/verification_test.go
```diff
@@ -104,7 +104,7 @@ func TestVerification_Expiration(t *testing.T) {
 
 func TestVerification_IsExpired(t *testing.T) {
 	tim, _ := time.Parse(time.RFC3339, "2021-03-16T04:19:57.592Z")
-	tim2, _ := time.Parse(time.RFC3339, "2022-03-16T04:19:57.592Z")
+	tim2 := time.Now().Add(time.Hour * 24)
```
rot1024, line 107: ok for now but also remember this way:

```go
var timeNow = time.Now

func hoge() {
  now := timeNow()
}
```

```go
func Test_hoge() {
	timeNow = func () time.Time {
		return time.Time{...}
	}
	defer timeNow = time.Now

	// test
}
```

NOTE: `t.Parallel()` is not supported

rot1024 (approved):

PR #84, opened 2021-11-29
feat: authentication system - mongo storage

yk-eukarya:
- Change auth storage into mongo db 
- Some refactoring on structure level


internal/infrastructure/mongo/mongodoc/auth_request.go
```diff
@@ -0,0 +1,116 @@
...
+		AuthorizedAt:  req.AuthorizedAt(),
+	}, reqID
+}
+
+func (d *AuthRequestDocument) Model() (*auth.Request, error) {
+	if d == nil {
+		return nil, nil
+	}
+
+	uuid, err := id.AuthRequestIDFrom(d.ID)
```
rot1024, line 88: Just to make sure you're not mistaken, generated IDs of pkg/id is not UUID, but U"L"ID (https://github.com/ulid/spec)

internal/usecase/interactor/auth.go
```diff
@@ -134,87 +135,81 @@ func (s *Storage) Health(_ context.Context) error {
...
+	if err := s.requests.Save(ctx, request); err != nil {
+		return nil, err
+	}
 	return request, nil
 }
 
-func (s *Storage) AuthRequestByID(_ context.Context, requestID string) (op.AuthRequest, error) {
+func (s *Storage) AuthRequestByID(ctx context.Context, requestID string) (op.AuthRequest, error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
```
rot1024, line 178: Lock is no longer needed. Delete the field also
```suggestion
```

internal/usecase/interactor/auth.go
```diff
@@ -1,4 +1,4 @@
-package appauth
```
rot1024, line 1: `interactor.Storage` is too abstract name and I don't know what the storage is in the interactor. `interactor.AuthStorage` is ok.

rot1024 (changes requested):
As a matter of fact, I thought it would be better to pack the implementation of authentication in `appauth` package, because in DDD, authentication is a subdomain, not a core domain.

However, your implementation is fine. The package is well separated, so it is unlikely to be a stumbling block for future refactoring.

Requests:

- Please implement `internal/infrastructure/memory/auth.go` also.

rot1024 (approved):

PR #88, opened 2021-12-10
feat: store auth cert in db

yk-eukarya:


internal/usecase/interactor/auth.go
```diff
@@ -74,8 +76,31 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
 			PostalCode:         cfg.DN.PostalCode,
 		}
 	}
+	c, err := config.Load(ctx)
+	if err != nil {
+		panic("Could not load auth config")
```
rot1024, line 81:
```suggestion
		log.Panicf("Could not load auth config: %s", err)
```

Note that I commented that NewAuthStorage should return errors instead of panicking

internal/usecase/interactor/auth.go
```diff
@@ -74,8 +76,31 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
...
+		keyBytes, certBytes, err = generateCert(name)
+		if err != nil {
+			panic("Could not generate raw cert")
+		}
+		c.Auth = &config2.Auth{
+			Key:  keyBytes,
+			Cert: certBytes,
+		}
+		err := config.Save(ctx, c)
+		if err != nil {
```
rot1024, line 98:
```suggestion
		if err := config.Save(ctx, c); err != nil {
```

internal/usecase/interactor/auth.go
```diff
@@ -90,28 +115,19 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
...
 
-func initKeys(name pkix.Name) (*rsa.PrivateKey, jose.SigningKey, jose.JSONWebKeySet) {
+func initKeys(keyBytes, certBytes []byte) (*rsa.PrivateKey, jose.SigningKey, jose.JSONWebKeySet) {
 
-	key, err := rsa.GenerateKey(rand.Reader, 2048)
-	if err != nil {
-		panic(err)
+	block, _ := pem.Decode(keyBytes)
+	if block == nil {
+		panic("failed to create the key")
```
rot1024, line 122:
```suggestion
	block, err := pem.Decode(keyBytes)
	if err != nil || block == nil {
		return nil, err
```

internal/usecase/interactor/auth.go
```diff
@@ -90,28 +115,19 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
 	}
 }
 
-func initKeys(name pkix.Name) (*rsa.PrivateKey, jose.SigningKey, jose.JSONWebKeySet) {
+func initKeys(keyBytes, certBytes []byte) (*rsa.PrivateKey, jose.SigningKey, jose.JSONWebKeySet) {
```
rot1024, line 118:
```suggestion
func initKeys(keyBytes, certBytes []byte) (*rsa.PrivateKey, jose.SigningKey, jose.JSONWebKeySet, error) {
```

internal/usecase/interactor/auth.go
```diff
@@ -57,7 +59,7 @@ var dummyName = pkix.Name{
 	PostalCode:         []string{"1"},
 }
 
-func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubject func(context.Context, string) (*user.User, error)) op.Storage {
+func NewAuthStorage(ctx context.Context, cfg *StorageConfig, request repo.AuthRequest, config repo.Config, getUserBySubject func(context.Context, string) (*user.User, error)) op.Storage {
```
rot1024, line 62:
```suggestion
func NewAuth(ctx context.Context, cfg *StorageConfig, request repo.AuthRequest, config repo.Config, getUserBySubject func(context.Context, string) (*user.User, error)) (op.Storage, error) {
```

Panic should be caused in one function; other functions should return errors, because we may want to change the behavior to avoid panicking, and if the panic calls are littered with them, it can be an obstacle to troubleshooting.

internal/infrastructure/mongo/config.go
```diff
@@ -24,7 +25,7 @@ func NewConfig(client *mongodoc.Client) repo.Config {
 
 func (r *configRepo) Load(ctx context.Context) (*config.Config, error) {
 	cfg := &config.Config{}
-	if err := r.client.Collection().FindOne(ctx, nil).Decode(cfg); err != nil {
+	if err := r.client.Collection().FindOne(ctx, bson.D{}).Decode(cfg); err != nil {
```
rot1024, line 28: I don't think this change is necessary if config can be successfully loaded before saving it, why is this change necessary?

rot1024 (changes requested):

internal/usecase/interactor/auth.go
yk-eukarya: `pem.Decode` dos not returns error as a second return value, check the docs [here](https://pkg.go.dev/encoding/pem#Decode)

rot1024 (changes requested):

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
 
-var upsert = true
+var upsert = false
 
 type configRepo struct {
 	client *mongodoc.ClientCollection
+	lockID *uuid.UUID
 }
 
+type Config struct {
```
rot1024, line 27:
```suggestion
type ConfigDoc struct {
```

It may confuse with `pkg/config.Config`

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
+		lockID := uuid.New()
+		r.lockID = &lockID
+	}
+
+	if err := r.client.Collection().FindOneAndUpdate(ctx,
+		bson.M{},
+		/*bson.M{"$or": bson.A{
+			bson.M{"lock": bson.M{"$exists": false}},
+			bson.M{"lock": bson.M{"$eq": r.lockID}},
+		}},*/
```
rot1024, line 72:
```suggestion
		bson.M{"$or": bson.A{
			bson.M{"lock": bson.M{"$exists": false}},
			bson.M{"lock": bson.M{"$eq": r.lockID}},
		}},
```

FindOneAndUpdate requires a filter. If it is not needs, actually `FindOne` is enough. In this case, FindOneAndUpdate is required to avoid conflicting acquisition of locks by other applications.

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
+	}
+	return &config.Config{
+		Migration: cfg.Migration,
+		Auth:      cfg.Auth,
+	}, nil
+}
+
+func (r *configRepo) Release(ctx context.Context) error {
+	if _, err := r.client.Collection().UpdateOne(ctx,
+		bson.M{},
```
rot1024, line 91:
```suggestion
func (r *configRepo) Release(ctx context.Context) error {
	if r.lockID == nil {
		return ErrConfigNotLocked // ErrSavingNotLockedConfig  can be renamed to
	}
	if _, err := r.client.Collection().UpdateOne(ctx,
		bson.M{"lock": bson.M{"$eq": r.lockID}},
```

Current code might unlock a locked document by others.

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
+		bson.M{"$and": bson.A{
+			bson.M{"lock": bson.M{"$exists": true}},
+			bson.M{"lock": bson.M{"$eq": r.lockID}},
+		}},
+		bson.M{"$set": cfg}, &options.UpdateOptions{
+			Upsert: &upsert,
+		}); err != nil {
 		return rerror.ErrInternalBy(err)
 	}
 	return nil
```
rot1024, line 116:
```suggestion
	r.lockID = nil
	return nil
```

This ensures the lock was completely unlocked. The next time the config is loaded, the new lock ID will be used.

internal/infrastructure/memory/config.go
```diff
@@ -17,14 +17,21 @@ func NewConfig() repo.Config {
...
 	r.lock.Lock()
 	defer r.lock.Unlock()
 
 	return r.data, nil
 }
 
-func (r *Config) Save(ctx context.Context, c *config.Config) error {
+func (r *Config) Release(_ context.Context) error {
+	r.lock.Lock()
+	defer r.lock.Unlock()
```
rot1024, line 29:
```suggestion
```

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-var upsert = true
+var upsert = false
 
 type configRepo struct {
 	client *mongodoc.ClientCollection
+	lockID *uuid.UUID
```
rot1024, line 24:
```suggestion
	lockID *uuid.UUID
	lock sync.Mutex
```

OK, mongo is now synced, but the local should be also synced. If multiple goroutines call the Load method at the same time, the lockID field can be conflicted. `r.lock.Lock(); defer r.lock.Unlock()` is required to each methods of configRepo.

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
+
+	for i := 1; i <= numOfTries; i++ {
+		cfg, err := r.loadFromDB(ctx)
+		if err == nil {
 			return cfg, nil
 		}
+		if !errors.Is(err, ErrLoadingLockedConfig) {
+			return nil, err
+		}
+		time.Sleep(time.Duration(rand.Intn(1000)) * time.Millisecond)
```
rot1024, line 53: With this code, the fastest duration could be 0 milliseconds. It is better to set a minimum duration.

```suggestion
		time.Sleep(time.Duration(rand.Intn(1000) + 1000) * time.Millisecond)
```

rot1024 (changes requested):
It's very nice that the locking related processes are hidden from outside of infrastructure/mongo.

internal/infrastructure/mongo/config.go
```diff
@@ -3,44 +3,114 @@ package mongo
...
 	"github.com/reearth/reearth-backend/pkg/config"
+	"github.com/reearth/reearth-backend/pkg/log"
 	"github.com/reearth/reearth-backend/pkg/rerror"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
-var upsert = true
+var upsert = false
```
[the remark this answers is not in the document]
yk-eukarya: If we keep upsert as true this may cause a duplication of the config document. I'm not sure can you confirm if is that ok?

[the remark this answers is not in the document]
rot1024: Upsert false is ok for config, but I worry about it affects to other repos. Separate variable: upsertFalse := false

internal/infrastructure/mongo/config.go
yk-eukarya: I did it this way so I can check if the document is not exists or locked! anyway the after this change I think we need to be sure that the database always have this document

rot1024 (changes requested):
Can you modify `mogno/migration/client.go` also to use config repo?

internal/infrastructure/mongo/config.go
rot1024: [See https://github.com/reearth/reearth-backend/pull/88#pullrequestreview-834816759](https://github.com/reearth/reearth-backend/pull/88#pullrequestreview-834816759)

internal/infrastructure/mongo/config.go
```diff
@@ -16,30 +22,103 @@ var upsert = true
...
-		Upsert: &upsert,
-	}); err != nil {
+	if r.lockID == nil {
+		return ErrConfigNotLocked
+	}
+	if _, err := r.client.Collection().UpdateOne(ctx,
+		bson.M{"lock": bson.M{"$eq": r.lockID}},
+		bson.M{"$set": cfg}, &options.UpdateOptions{
+			Upsert: &upsert,
+		}); err != nil {
```
rot1024, line 121: Refactoring

```suggestion
	if _, err := r.client.Collection().UpdateOne(
		ctx,
		bson.M{"lock": bson.M{"$eq": r.lockID}},
		bson.M{"$set": cfg},
		(&options.UpdateOptions{}).SetUpsert(true),
	); err != nil {
```

internal/infrastructure/mongo/config.go
```diff
@@ -16,30 +22,103 @@ var upsert = true
...
+		r.lockID = &lockID
+	}
+
+	if err := r.client.Collection().FindOneAndUpdate(ctx,
+		bson.M{"$or": bson.A{
+			bson.M{"lock": bson.M{"$exists": false}},
+			bson.M{"lock": bson.M{"$eq": r.lockID}},
+		}},
+		bson.M{"$set": bson.M{"lock": r.lockID}},
+	).Decode(cfg); err != nil {
```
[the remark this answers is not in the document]
yk-eukarya: Using upsert leads into config duplication in db, I tried this suggestion and it cause a duplication when tiring to load a locked config.
anyway I updated the implementation but this assumes the the config should be in the db other wise it will not created automatically

internal/usecase/interactor/auth.go
```diff
@@ -74,46 +78,68 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
 			PostalCode:         cfg.DN.PostalCode,
 		}
 	}
+	c, err := config.LockAndLoad(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("Could not load auth config: %w\n", err)
+	}
+	defer func() {
+		if err := config.Unlock(ctx); err == nil {
+			log.Errorf("Could not release config lock: %s\n", err)
```
rot1024, line 87:
```suggestion
			log.Errorf("auth: Could not release config lock: %s\n", err)
```

pkg/config/config.go
```diff
@@ -4,6 +4,12 @@ import "sort"
 
 type Config struct {
 	Migration int64
+	Auth      *Auth
+}
+
+type Auth struct {
+	Cert []byte
+	Key  []byte
```
rot1024, line 12:
```suggestion
	Cert string
	Key  string
```

Storing certificates as strings even within MongoDB makes it easier for infrastructure engineers to troubleshoot certificate issues.

rot1024 (changes requested):
Please do not forget modifying https://github.com/reearth/reearth-backend/blob/8b0c250b1deb5349f9aba851737e452291110bb3/internal/infrastructure/memory/config.go#L11

internal/usecase/interactor/auth.go
```diff
@@ -74,46 +79,72 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
...
+			return nil, err
+		}
+	} else {
+		keyBytes, certBytes, err = generateCert(name)
+		if err != nil {
+			return nil, fmt.Errorf("Could not generate raw cert: %w\n", err)
+		}
+		c.Auth = &config2.Auth{
+			Key:  base64.StdEncoding.EncodeToString(keyBytes),
+			Cert: base64.StdEncoding.EncodeToString(certBytes),
```
rot1024, line 107:
```suggestion
			Key: string(keyBytes),
			Cert: string(certBytes),
```

internal/usecase/interactor/auth.go
```diff
@@ -74,46 +79,72 @@ func NewAuthStorage(cfg *StorageConfig, request repo.AuthRequest, getUserBySubje
...
+	}()
+
+	var keyBytes, certBytes []byte
+	if c.Auth != nil {
+		if keyBytes, err = base64.StdEncoding.DecodeString(c.Auth.Key); err != nil {
+			return nil, err
+		}
+		if certBytes, err = base64.StdEncoding.DecodeString(c.Auth.Cert); err != nil {
+			return nil, err
+		}
```
rot1024, line 99:
```suggestion
		keyBytes = []byte(c.Auth.Key)
		certBytes = []byte(c.Auth.Cert)
```

rot1024 (changes requested):
PEM is human-readable and portable format and it can be casted between string and []byte. Using base64 is only worsens performance and does not make troubleshooting (specifying certificates directly in MongoDB's GUI) easier, so it doesn't make much sense.

rot1024 (approved):

PR #95, opened 2021-12-27
feat: add "clamp to ground" option to file primitive

HideBa:
# Overview


pkg/builtin/manifest.yml
```diff
@@ -1144,6 +1144,11 @@ extensions:
                   label: GeoJSON / TopoJSON
                 - key: czml
                   label: CZML
+            - id: clampToGround
+              type: bool
+              title: Clamp to ground
+              description: Option to clamp to ground or not.
```
KaWaite, line 1150:
```suggestion
              description: Option to attach the primitive to the ground (keeps primitive on top of the map even with terrain enabled).
```

KaWaite: Please check that the description is accurate!

KaWaite (approved):
Updated English. Check accuracy of my updated description before committing

pkg/builtin/manifest_ja.yml
```diff
@@ -510,6 +510,9 @@ extensions:
               kml: KML
               geojson: GeoJSON / TopoJSON
               czml: CZML
+          clampToGround:
+            title: 地表に固定
+            description: 外部データを地表に合わせて固定します。
```
rot1024, line 515:
```suggestion
            description: 地表の凹凸に沿うようにポリゴンなどのデータを表示します。
```

rot1024 (changes requested):

PR #102, opened 2022-01-27
feat: auth client support multi issuer

yk-eukarya:
# Overview

### DIFF
--- a/internal/app/jwt.go
+++ b/internal/app/jwt.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"time"
 
@@ -22,13 +23,13 @@ const (
 
 type MultiValidator []*validator.Validator
 
-func NewMultiValidator(providers []identityProvider) (MultiValidator, error) {
+func NewMultiValidator(providers []AuthConfig) (MultiValidator, error) {
 	validators := make([]*validator.Validator, 0, len(providers))
 	for _, p := range providers {
 
 		issuerURL, err := url.Parse(p.ISS)
 		if err != nil {
-			log.Fatalf("failed to parse the issuer url: %v", err)
+			return nil, fmt.Errorf("failed to parse the issuer url: %w", err)
 		}
 
 		provider := jwks.NewCachingProvider(issuerURL, time.Duration(*p.TTL)*time.Minute)
--- a/internal/app/jwt.go
+++ b/internal/app/jwt.go
@@ -20,23 +20,51 @@ const (
 	contextUser     contextKey = "reearth_user"
 )
 
+type MultiValidator []*validator.Validator
+
+func NewMultiValidator(providers []identityProvider) (MultiValidator, error) {
+	validators := make([]*validator.Validator, 0, len(providers))
+	for _, p := range providers {
+
+		issuerURL, err := url.Parse(p.ISS)
+		if err != nil {
+			log.Fatalf("failed to parse the issuer url: %v", err)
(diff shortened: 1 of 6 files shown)


internal/app/jwt.go
```diff
@@ -2,223 +2,74 @@ package app
...
-	var res struct {
-		Jwks []JSONWebKeys `json:"keys"`
-	}
+	provider := jwks.NewCachingProvider(issuerURL, 60*time.Minute)
 
+	// Set up the validator.
+	jwtValidator, err := validator.New(
+		provider.KeyFunc,
+		validator.RS256,
+		cfg.Config.AuthClient.ISS[0],
```
rot1024, line 37: This implementation does not support multiple isses; either change the PR title to "Update go-jwt-middleware to v2" and complete the PR once with just the update, or implement multiple iss support. You will probably have to write your own validator.

internal/app/jwt.go
```diff
@@ -2,223 +2,74 @@ package app
...
-				}
+				ctx = context.WithValue(ctx, contextAuth0Sub, claims.RegisteredClaims.Subject)
+
+				/* this claim is sent from auth0
 				if user, ok := claims["https://reearth.io/user_id"].(string); ok {
 					ctx = context.WithValue(ctx, contextUser, user)
-				}
-				ctx = context.WithValue(ctx, contextAuth0AccessToken, userProfile.Raw)
-			}
+				}*/
```
rot1024, line 65: Please delete comments that simply comment out the code, as they cause confusion. We use Git, so we can browse past code at any time.

internal/app/jwt.go
```diff
@@ -2,223 +2,74 @@ package app
...
-		SigningMethod:       jwt.SigningMethodRS256,
-		// Make jwtmiddleware return an error object by not writing ErrorHandler to ResponseWriter
-		ErrorHandler: func(w http.ResponseWriter, req *http.Request, err string) {},
-		ValidationKeyGetter: func(token *jwt.Token) (interface{}, error) {
-			cert, err := getPemCert(token, jwksURL, jwks)
-			if err != nil {
-				log.Errorf("jwt: %s", err)
-				return nil, err
+				// this one is not used!
+				// ctx = context.WithValue(ctx, contextAuth0AccessToken, userProfile.Raw)
```
rot1024, line 68: Delete code

rot1024 (changes requested):
- It seems that some modules are no longer needed. Please run `go mod tidy`.
- Please make sure that the behavior of this source code is not corrupted. https://github.com/reearth/reearth-backend/blob/main/internal/adapter/gql/context.go
- To implement our own validator to support multiple iss is not hard to use delegate pattern:

```go
type MultiValidator [[jwtmiddleware.Validator

func NewMultiValidator(/*....*/, iss []string, aud string) (MultiValidator, error) {
	vv := make([]jwtmiddleware.Validator, 0, len(iss))
	for _, i := range iss {
		v, err := validator.New(
			// ...,
			i,
			aud,
			// ...
		)
		if err != nil {
			return nil, err
		}
		vv = append(vv, v)
	}
	return Validator(vv, jwtmiddleware.Validator), nil
}

func (vv MultiValidator) ValidateToken(ctx context.Context, tokenString string) (res interface{}, err error) {
	for _, v := range vv {
		res, err = v(ctx, tokenString)
		if err == nil {
			return
		}
	}
	return
}
```

rot1024 (changes requested):

rot1024 (approved):

PR #116, opened 2022-02-18
feat: validate password criteria

mimoham24:
# Overview

### DIFF
--- a/pkg/user/user.go
+++ b/pkg/user/user.go
@@ -2,8 +2,9 @@ package user
 
 import (
 	"errors"
-	"golang.org/x/crypto/bcrypt"
 	"unicode"
+
+	"golang.org/x/crypto/bcrypt"
 
 	"golang.org/x/text/language"
 )
--- a/pkg/user/user.go
+++ b/pkg/user/user.go
@@ -2,8 +2,8 @@ package user
 
 import (
 	"errors"
-
 	"golang.org/x/crypto/bcrypt"
+	"unicode"
 
 	"golang.org/x/text/language"
 )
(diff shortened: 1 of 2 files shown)


pkg/user/user.go
```diff
@@ -193,3 +202,31 @@ func (u *User) SetPasswordReset(pr *PasswordReset) {
...
+
+func validatePassword(pass string) error {
+	var hasNum, hasUpper, hasLower bool
+	for _, c := range pass {
+		switch {
+		case unicode.IsNumber(c):
+			hasNum = true
+		case unicode.IsUpper(c):
+			hasUpper = true
+		case unicode.IsLetter(c) || c == ' ':
```
yk-eukarya, line 214:
```suggestion
		case unicode.IsLower(c) || c == ' ':
```

yk-eukarya (approved):
LGTM

pkg/user/user.go
```diff
@@ -151,6 +156,10 @@ func (u *User) ClearAuths() {
 }
 
 func (u *User) SetPassword(pass string) error {
+	err := validatePassword(pass)
+	if err != nil {
+		return err
+	}
```
rot1024, line 162:
```suggestion
	if err := validatePassword(pass); err != nil {
		return err
	}
```

rot1024 (changes requested):
Uppercase and lowercase concept of unicode are not limited to alphabet but other language characters. Is that expected?

- https://www.compart.com/en/unicode/category/Lu
- https://www.ibm.com/docs/en/i/7.2?topic=tables-unicode-lowercase-uppercase-conversion-mapping-table

mimoham24:
@rot1024 after investigations it's ok to input a password with non-Latin characters 
the only concern is that it might cause encoding and decoding problems
so, after a meeting with @yk-eukarya we thought that having a test case with non-Latin characters will make sure that `crypto/bcrypt` can successfully encode and decode non-Latin passwords

rot1024:
crypto/bcrypt should work. It is just a binary, so it doesn't matter at all what language the password is written in to bcrypt!

rot1024 (approved):

PR #120, opened 2022-03-02
refactor: refactor signup code

mimoham24:
# Overview
refactor signup code

## What I've done
- separate auth0 and auth system signup logic
- handle not verified user


internal/app/config.go
```diff
@@ -46,7 +46,7 @@ type Auth0Config struct {
 }
 
 type AuthSrvConfig struct {
-	Domain string `default:"http://localhost:8080"`
+	Domain string `default:"https://localhost:8080"`
```
rot1024, line 49:
```suggestion
	Domain string `default:"http://localhost:8080"`
```

HTTPS is not suitable for local servers because it requires self-signed certificates

internal/usecase/interactor/user.go
```diff
@@ -241,6 +169,109 @@ func (i *User) Signup(ctx context.Context, inp interfaces.SignupParam) (u *user.
...
+
+	tx, err := i.transaction.Begin()
+	if err != nil {
+		return nil, "", "", nil, nil, err
+	}
+	defer func() {
+		if err2 := tx.End(ctx); err == nil && err2 != nil {
+			err = err2
+		}
+	}()
```
rot1024, line 191: There are duplication of this and line 228. 
- Begin transaction outside this function.
- There is no need to accept and return `repo.Transaction` and `repo.Tx`. 

Fix `auth0Signup` also.

internal/usecase/interactor/user.go
```diff
@@ -241,6 +169,109 @@ func (i *User) Signup(ctx context.Context, inp interfaces.SignupParam) (u *user.
...
+		}
+	}
+
+	return tx, *inp.Name, *inp.Email, nil, nil, nil
+}
+
+func (i *User) auth0Signup(ctx context.Context, inp interfaces.SignupParam, tx repo.Tx) (repo.Tx, string, string, *user.Auth, error) {
+	if i.signupSecret != "" && *inp.Secret != i.signupSecret {
+		return nil, "", "", nil, interfaces.ErrSignupInvalidSecret
+	}
```
rot1024, line 222: Move also this validation out of the function to move codes to start transaction.

internal/usecase/interactor/user.go
```diff
@@ -241,6 +169,109 @@ func (i *User) Signup(ctx context.Context, inp interfaces.SignupParam) (u *user.
...
+func (i *User) authSystemSignup(ctx context.Context, inp interfaces.SignupParam, tx repo.Tx) (repo.Tx, string, string, *user.User, *user.Team, error) {
+	if *inp.Name == "" {
+		return nil, "", "", nil, nil, interfaces.ErrSignupInvalidName
+	}
+	if _, err := mail.ParseAddress(*inp.Email); err != nil {
+		return nil, "", "", nil, nil, interfaces.ErrInvalidUserEmail
+	}
+	if *inp.Password == "" {
+		return nil, "", "", nil, nil, interfaces.ErrSignupInvalidPassword
+	}
```
rot1024, line 181: Move also this validation out of the function to move codes to start transaction.

rot1024 (changes requested):

internal/app/config.go
mimoham24: sorry it's forgotten 
I can't run the server locally without HTTPS

rot1024: Could you please elaborate on what the situation is?

internal/usecase/interactor/user.go
mimoham24: I wanted to start the transaction just before the query code

rot1024: Current:

pseudo code

```
if auth0 {
  start tx // duplicated
  auth0Signup()
} else if authSystem {
  start tx // duplicated
  authSystemSignup()
}
```

Goal:

pseudo code
```
if !auth0 && !signup {
  return
}
start tx
if auth0 {
  auth0Signup()
} else if authSystem {
  authSystemSignup()
}
```

rot1024 (changes requested):
- There still seems to be a duplication. You might try the following
- Remame `authSystemSignup` to `reearthSignup` to clarify meaning
- In fact, for OIDC compliance, user information must be obtained from `/userinfo`, one of the OIDC endpoints, rather than from the Auth0 API. Please do so. Also, the function name `auth0Signup` is no longer appropriate due to its implementation; it should be changed to `oidcSignup`.

Current:

pseudo code

```
if auth0 {
  start tx // duplicated
  auth0Signup()
} else if authSystem {
  start tx // duplicated
  authSystemSignup()
}
```

Goal:

pseudo code
```
if !auth0 && !signup {
  return
}
start tx
if auth0 {
  auth0Signup()
} else if authSystem {
  authSystemSignup()
}
```

internal/usecase/interactor/user.go
```diff
@@ -572,18 +597,31 @@ func (i *User) CreateVerification(ctx context.Context, email string) error {
...
+
+	vr := user.NewVerification()
+	u.SetVerification(vr)
 	err = i.userRepo.Save(ctx, u)
 	if err != nil {
 		return err
 	}
 
+	var TextOut, HTMLOut bytes.Buffer
+	link := "localhost:3000/user-verification-token=" + vr.Code()
```
yk-eukarya, line 609:
```suggestion
	link := "localhost:3000/?user-verification-token=" + vr.Code()
```

internal/infrastructure/mailer/smtp.go
```diff
@@ -67,6 +69,7 @@ func (m *message) encodeContent() (string, error) {
 func (m *message) encodeMessage() ([]byte, error) {
 	buf := bytes.NewBuffer(nil)
 	buf.WriteString(fmt.Sprintf("Subject: %s\n", m.subject))
+	buf.WriteString(fmt.Sprintf("from: %s\n", m.from))
```
rot1024, line 72:
```suggestion
	buf.WriteString(fmt.Sprintf("From: %s\n", m.from))
```

rot1024 (approved):

PR #139, opened 2022-05-08
test: add Mongo Asset's `FindByID` unit testing

KeisukeYamashita:
# Overview

I have added initial unit testing for the MongoDB based on this issue for "small start" 👉  https://github.com/reearth/reearth/issues/273.
See details (requirements) in that issue.

I implemented one for a trial to avoid conflicts and decide how we are going to test this methods.

## What I've done

I have added unit test for the Asset repository's `FindByID` method.

## What I haven't done

Add unit tests for all methods for Assets or other repositories.
This is because I wanted to "start small" since we need to discuss the policy and the implementation design first so that we don't waste time.

## How I tested

The testings are done by the CI GitHub Actions workflow.
This can be also tested locally.

```shell
# Skip testing
go test -timeout 30s -run '^TestFindByID$' github.com/reearth/reearth-backend/internal/infrastructure/mongo

# Test (after launching MongoDB instance in your local environment)
REEARTH_DB=mongodb://localhost go test -timeout 30s -run '^TestFindByID$' github.com/reearth/reearth-backend/internal/infrastructure/mongo
```

* How the testing(s) should be done
    * Table testing
    * Variable namings
    * etc 

## Memo

Please feel free to give me any feedback as this is the starting point of adding unit tests to the repository layer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,11 @@ jobs:
   ci:
     name: CI
     runs-on: ubuntu-latest
+    services:
+      mongo:
+        image: mongo:4.4-focal
+        ports:
+          - 27017:27017
     steps:
       - name: set up
         uses: actions/setup-go@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,8 @@ jobs:
           args: --timeout=10m
       - name: test
         run: go test ./... -v -race -coverprofile=coverage.txt -covermode=atomic -timeout 10m
+        env:
+          REEARTH_DB: mongodb://localhost
       - name: Send coverage report
         uses: codecov/codecov-action@v2
         with:
--- a/internal/infrastructure/mongo/asset_test.go
+++ b/internal/infrastructure/mongo/asset_test.go
@@ -2,27 +2,15 @@ package mongo
 
 import (
 	"context"
-	"os"
 	"testing"
 	"time"
 
-	"github.com/reearth/reearth-backend/internal/infrastructure/mongo/mongodoc"
 	"github.com/reearth/reearth-backend/pkg/asset"
 	"github.com/reearth/reearth-backend/pkg/id"
 	"github.com/stretchr/testify/assert"
-	"go.mongodb.org/mongo-driver/mongo"
-	"go.mongodb.org/mongo-driver/mongo/options"
-	"go.mongodb.org/mongo-driver/x/mongo/driver/uuid"
 )
 
 func TestFindByID(t *testing.T) {
-	// Skip unit testing if "REEARTH_DB" is not configured
-	// See details: https://github.com/reearth/reearth/issues/273
-	db := os.Getenv("REEARTH_DB")
-	if db == "" {
-		return
-	}
-
 	tests := []struct {
 		Name     string
 		Expected struct {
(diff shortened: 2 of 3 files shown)


internal/infrastructure/mongo/asset_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestFindByID(t *testing.T) {
+	// Skip unit testing if "REEARTH_DB" is not configured
+	// See details: https://github.com/reearth/reearth/issues/273
+	db := os.Getenv("REEARTH_DB")
+	if db == "" {
+		return
+	}
+
+	type a struct {
```
KeisukeYamashita, line 26: Note 📖 

I locally named `a` but if you have preference, please let me know.
(Also, we can rename the `asset` in the `import` to other name )

internal/infrastructure/mongo/asset_test.go
```diff
@@ -0,0 +1,108 @@
...
+				Team(tc.Expected.Asset.team).
+				Name(tc.Expected.Asset.name).
+				Size(tc.Expected.Asset.size).
+				URL(tc.Expected.Asset.url).
+				CreatedAt(tc.Expected.Asset.createdAt).
+				Build()
+
+			assert.NoError(t, err)
+
+			database, _ := uuid.New()
```
KeisukeYamashita, line 83: Note 📖 

I used the Mongo's (`go.mongodb.org/mongo-driver/x/mongo/driver/uuid`) `uuid` generator as the database name instead of the Google ones (`github.com/google/uuid`) which are more popular. This is because since we use Mongo already, we won't need to add additional packages.

.github/workflows/ci.yml
```diff
@@ -8,6 +8,11 @@ jobs:
   ci:
     name: CI
     runs-on: ubuntu-latest
+    services:
+      mongo:
+        image: mongo:4.4-focal
```
KeisukeYamashita, line 13: Note 📖 

Adjusted with the current MongoDB version spec declared in Docker compose config.

https://github.com/reearth/reearth-backend/blob/83f9b135f57e74d55361b0a51bd6ce1ce29b16c6/docker-compose.yml#L20

KeisukeYamashita:
TODO ✔️ 

* Need to add the `REEARTH_DB` env for testing in the CI

rot1024 (commented):
Thank you! How about creating a helper function like this one?

```go
func connect(t *testing.T) func() (*mongodoc.Client, func()) {
	t.Helper()

	// Skip unit testing if "REEARTH_DB" is not configured
	// See details: https://github.com/reearth/reearth/issues/273
	db := os.Getenv("REEARTH_DB")
	if db == "" {
		t.SkipNow()
		return nil
	}

	c, _ := mongo.Connect(
		context.Background(),
		options.Client().
			ApplyURI(db).
			SetConnectTimeout(time.Second*10),
	)

	return func() {
		database, _ := uuid.New()
		client := mongodoc.NewClient(string(database[:]), c)
		return client, func() {
			_ = c.Database(string(database[:])).Drop(context.Background())
		}
	}
}
```

Usage:

```go
func TestFindByID(t *testing.T) {
	// ...
	initDB := connect(t)
 
	for _, tc := range tests {
		tc := tc
		t.Run(tc.Name, func() {
			t.Paralell()
			client, dropDB := initDB()
			defer dropDB()

			repo := NewAsset(client)	  
			ctx := context.Background()
			err := repo.Save(ctx, want)
			assert.NoError(t, err)
	  
			got, err := repo.FindByID(ctx, want.ID())
			// ...
		})
	}
}
```

internal/infrastructure/mongo/asset_test.go
KeisukeYamashita: Ah, I didn't noticed the `New()` function... Thank you for pointing it out! 👉 Updated 👉 https://github.com/reearth/reearth-backend/pull/139/commits/06278f8c9f22c5237f9574d9f23ef4d8e00372aa

rot1024 (approved):

PR #151, opened 2022-05-25
refactor: declarative description of use case structure (asset only)

rot1024:
Due to the low readability of use cases, the following statements can be made more declarative:

- Whether the usecase uses a transaction
- Whether the usecase denies users who cannot read/write the specific scene and team

### DIFF
--- a/internal/usecase/interactor/asset.go
+++ b/internal/usecase/interactor/asset.go
@@ -32,8 +32,8 @@ func (i *Asset) Fetch(ctx context.Context, assets []id.AssetID, operator *usecas
 func (i *Asset) FindByTeam(ctx context.Context, tid id.TeamID, keyword *string, sort *asset.SortType, p *usecase.Pagination, operator *usecase.Operator) ([]*asset.Asset, *usecase.PageInfo, error) {
 	return Run2(
 		ctx, operator, i.repos,
-		Usecase().CanReadTeams(tid),
-		func(ctx context.Context) ([]*asset.Asset, *usecase.PageInfo, error) {
+		Usecase().WithReadableTeams(tid),
+		func() ([]*asset.Asset, *usecase.PageInfo, error) {
 			return i.repos.Asset.FindByTeam(ctx, tid, repo.AssetFilter{
 				Sort:       sort,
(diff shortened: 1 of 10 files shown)


internal/usecase/interactor/usecase.go
```diff
@@ -0,0 +1,121 @@
...
+	writableTeams  id.TeamIDList
+	readableScenes id.SceneIDList
+	writableScenes id.SceneIDList
+}
+
+func Usecase() *uc {
+	return &uc{}
+}
+
+func (u *uc) CanReadTeams(ids ...id.TeamID) *uc {
```
yk-eukarya, line 24: @rot1024 
`Can` in the method name gives an impression that this method will return a bool value!
I think using a name like `WithReadableTeams` would be better

internal/usecase/interactor/common.go
```diff
@@ -46,6 +45,7 @@ func NewContainer(r *repo.Container, g *gateway.Container, config ContainerConfi
 	}
 }
 
+// common will be decrepated. Please use the Usecase function instead.
```
yk-eukarya, line 48:
```suggestion
// Deprecated: common will be deprecated. Please use the Usecase function instead.
```
using the `Deprecated:` at the beginning of the comment is a standard mentioned in [godocs](https://go.dev/blog/godoc), also it helps the IDE to give a better experience

internal/usecase/interactor/asset.go
```diff
@@ -31,95 +30,66 @@ func (i *Asset) Fetch(ctx context.Context, assets []id.AssetID, operator *usecas
...
+			}
+
+			return asset.New().
+				NewID().
+				Team(inp.TeamID).
+				Name(path.Base(inp.File.Path)).
+				Size(inp.File.Size).
+				URL(url.String()).
+				Build()
+		})
```
yk-eukarya, line 69: Extracting the logic as a private method would be more readable

rot1024: I don't think it would be that effective because the amount of code would increase...

yk-eukarya (approved):

PR #154, opened 2022-05-29
feat: add totalCount field to DatasetSchema type of GraphQL schema

keiya01:
# Overview

I fixed reearth/reearth#267.

## What I've done

- add `CountBySchema` method to count dataset by schema id.
- add totalCount resolver to DatasetSchema struct

## How I tested

1. run reearth-backend and reearth-web
2. add dataset
3. check dataset count

### DIFF
--- a/internal/infrastructure/memory/dataset.go
+++ b/internal/infrastructure/memory/dataset.go
@@ -87,6 +87,21 @@ func (r *Dataset) FindBySchema(ctx context.Context, id id.DatasetSchemaID, p *us
 	), nil
 }
 
+func (r *Dataset) CountBySchema(ctx context.Context, id id.DatasetSchemaID) (int, error) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	n := dataset.List{}
+	for _, dataset := range r.data {
+		if dataset.Schema() == id {
+			if r.f.CanRead(dataset.Scene()) {
+				n = append(n, dataset)
+			}
+		}
+	}
+	return len(n), nil
+}
+
 func (r *Dataset) FindBySchemaAll(ctx context.Context, id id.DatasetSchemaID) (dataset.List, error) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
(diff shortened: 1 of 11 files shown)


keiya01:
I ran `go generate ./...` to generate some gql code, but this generated too many code.
Is there another way to generate gql code?

[chore: generate gql](https://github.com/reearth/reearth-backend/pull/154/commits/37eb6bbe9d04e084a062ae5c94c473c74c9a56d4)

internal/infrastructure/memory/dataset.go
```diff
@@ -87,6 +87,21 @@ func (r *Dataset) FindBySchema(ctx context.Context, id id.DatasetSchemaID, p *us
...
+
+	n := dataset.List{}
+	for _, dataset := range r.data {
+		if dataset.Schema() == id {
+			if r.f.CanRead(dataset.Scene()) {
+				n = append(n, dataset)
+			}
+		}
+	}
+	return len(n), nil
```
rot1024, line 102:
```suggestion
	n := 0
	for _, dataset := range r.data {
		if dataset.Schema() == id {
			if r.f.CanRead(dataset.Scene()) {
				n++
			}
		}
	}
	return n, nil
```

internal/infrastructure/mongo/dataset.go
```diff
@@ -69,6 +69,16 @@ func (r *datasetRepo) FindBySchema(ctx context.Context, schemaID id.DatasetSchem
 	}, pagination)
 }
 
+func (r *datasetRepo) CountBySchema(ctx context.Context, id id.DatasetSchemaID) (int, error) {
+	res, err := r.client.Count(ctx, bson.M{
+		"schema": id.String(),
+	})
```
rot1024, line 75:
```suggestion
	res, err := r.client.Count(ctx, r.readFilter(bson.M{
		"schema": id.String(),
	}))
```

Filter queries to prevent access to documents that should not be visible to the current user.

rot1024 (changes requested):

PR #155, opened 2022-06-02
fix: assets are not saved when files are uploaded

rot1024:
# Overview

## What I've done

- Fix an issue that assets are not saved when files are uploaded
- Add a unit test for asset usecase partially
- Refactor `memory.New`
- Change `memory.Transaction.committed` behavior


yk-eukarya (approved):
LGTM

PR #159, opened 2022-06-17
test: unit test for mongo auth request repo

yuuri111:
# Overview
Test: mongo auth request

## What I've done
add unittest of FindByID in auth_request.go

### DIFF
--- a/internal/infrastructure/mongo/auth_request_test.go
+++ b/internal/infrastructure/mongo/auth_request_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_authRequestRepo_FindByID(t *testing.T) {
+func TestAuthRequestRepo(t *testing.T) {
 	tests := []struct {
 		Name     string
 		Expected struct {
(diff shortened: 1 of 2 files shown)


internal/infrastructure/mongo/auth_request_test.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/caos/oidc/pkg/oidc"
+	"github.com/reearth/reearth-backend/pkg/auth"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_authRequestRepo_FindByID(t *testing.T) {
```
rot1024, line 12:
```suggestion
func TestAuthRequestRepo(t *testing.T) {
```

Collect name is `TestXXX_YYY`, not `Test_XXX_YYY`, but this test includes unit tests for `Save` method also. So I suggest that name.

rot1024 (approved):

internal/infrastructure/mongo/auth_request_test.go
yuuri111: @rot1024 
Thank you, I fixed it.

PR #160, opened 2022-07-15
fix: remove property field issue

yk-eukarya:
# Overview

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 
 # Editor
 .idea
+.vscode
 
 # reearth
 /dist
--- a/pkg/property/property.go
+++ b/pkg/property/property.go
@@ -268,6 +268,8 @@ func (p *Property) RemoveField(ptr *Pointer) {
 
 	if group := ToGroup(p.Item(ptr)); group != nil {
 		group.RemoveField(fid)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -14,6 +14,7 @@
 
 # Editor
 .idea
+.vscode
```
rot1024, line 17:
```suggestion
```

pkg/property/property.go
```diff
@@ -268,6 +268,8 @@ func (p *Property) RemoveField(ptr *Pointer) {
 
 	if group := ToGroup(p.Item(ptr)); group != nil {
 		group.RemoveField(fid)
+	} else if _, groupList := p.GroupAndList(ptr); groupList != nil {
+		groupList.RemoveFields(ptr)
```
rot1024, line 272:
```suggestion
	item := p.Item(ptr)
	if group := ToGroup(item); group != nil {
		group.RemoveField(fid)
	} else if groupList := ToList(item); groupList != nil {
		groupList.RemoveFields(ptr)
```

rot1024 (changes requested):

rot1024 (approved):

